Issue #8, opened 2021-09-08
[BUG][Opensearch] Invalid volume permissions plus securityConfigSecrets


alborotogarcia:
**Describe the bug**
Failed to mount multiple secrets at once from securityConfigSecrets.config.data plus volume permissions 
**To Reproduce**
1. Having set
```securityConfig:
  enabled: true
  path: "/usr/share/opensearch/plugins/opensearch-security/securityconfig"
  actionGroupsSecret:
  configSecret:
  internalUsersSecret:
  rolesSecret:
  rolesMappingSecret:
  tenantsSecret:
  #The following option simplifies securityConfig by using a single secret and specifying the respective secrets in the corresponding files instead of creating different secrets for config,internal users, roles, roles mapping and tenants
  #Note that this is an alternative to the above secrets and shouldn't be used if the above secrets are used
  config:
    securityConfigSecret: mysecret
    data:
         config.yml: |-
```
Returns the following:
```Error: YAML parse error on opensearch/templates/securityconfig.yaml: error converting YAML to JSON: yaml: invalid map key: map[interface {}]interface {}{".Chart.Name":interface {}(nil)}
helm.go:88: [debug] error converting YAML to JSON: yaml: invalid map key: map[interface {}]interface {}{".Chart.Name":interface {}(nil)}```

After applying a few fixes at opensearch/templates/securityconfig.yaml

It shows the following:
```opensearch-cluster-master-1 opensearch [2021-09-08T05:44:38,943][INFO ][o.o.p.PluginsService     ] [opensearch-cluster-master-1] loaded plugin [opensearch-sql]
opensearch-cluster-master-1 opensearch [2021-09-08T05:44:39,058][INFO ][o.o.s.OpenSearchSecurityPlugin] [opensearch-cluster-master-1] Disabled https compression by default to mitigate BREACH attacks. You can enable it by setting 'http.compression: true' in opensearch.yml
opensearch-cluster-master-1 opensearch [2021-09-08T05:44:39,613][ERROR][o.o.b.OpenSearchUncaughtExceptionHandler] [opensearch-cluster-master-1] uncaught exception in thread [main]
opensearch-cluster-master-1 opensearch org.opensearch.bootstrap.StartupException: OpenSearchException[failed to bind service]; nested: AccessDeniedException[/usr/share/opensearch/data/nodes];
```
Due to fsgroup permissions..

**Expected behavior**

**Chart Name**
Specify the Chart which is affected?
Opensearch
[...]

PR #9, opened 2021-09-08
fix securityConfigSecrets.config.data secrets mount plus permissions

alborotogarcia:
### Description
[Describe what this change achieves]
Check volume mount permissions to 1000
Fix securityConfigSecrets mounts
### Issues Resolved
[List any issues this PR will resolve]
 #8 
### Check List

### DIFF
--- a/charts/opensearch/templates/securityconfig.yaml
+++ b/charts/opensearch/templates/securityconfig.yaml
@@ -5,9 +5,9 @@ metadata:
   name: {{ .Values.securityConfig.config.securityConfigSecret }}
   namespace: {{ .Release.Namespace }}
   labels:
-    app: "{{ .Chart.Name }}"
-    release: {{ .Release.Name | quote }}
-    heritage: {{ .Release.Service | quote }}
+    app: {{ .Chart.Name }}
+    release: {{ .Release.Name }}
+    heritage: {{ .Release.Service }}
 type: Opaque
 data:
   {{- range $key, $val := .Values.securityConfig.config.data }}
--- a/charts/opensearch/templates/securityconfig.yaml
+++ b/charts/opensearch/templates/securityconfig.yaml
@@ -5,9 +5,9 @@ metadata:
   name: {{ .Values.securityConfig.config.securityConfigSecret }}
   namespace: {{ .Release.Namespace }}
   labels:
-    app: {{ .Chart.Name }}
-    release: {{ .Release.Name }}
-    heritage: {{ .Release.Service }}
+    app: "{{ .Chart.Name }}"
+    release: {{ .Release.Name | quote }}
+    heritage: {{ .Release.Service | quote }}
 type: Opaque
 data:
   {{- range $key, $val := .Values.securityConfig.config.data }}
--- a/charts/opensearch/templates/statefulset.yaml
+++ b/charts/opensearch/templates/statefulset.yaml
@@ -211,6 +211,7 @@ spec:
       {{- if .Values.hostAliases }}
       hostAliases: {{ toYaml .Values.hostAliases | nindent 8 }}
       {{- end }}
+      {{- if or (.Values.extraInitContainers) (.Values.keystore) (.Values.persistence.enabled) }}
       initContainers:
 {{- if .Values.persistence.enabled }}
       - name: fsgroup-volume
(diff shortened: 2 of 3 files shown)


charts/opensearch/templates/securityconfig.yaml
```diff
@@ -5,9 +5,9 @@ metadata:
   name: {{ .Values.securityConfig.config.securityConfigSecret }}
   namespace: {{ .Release.Namespace }}
   labels:
-    app: {{ .Chart.Name }}
-    release: {{ .Release.Name }}
-    heritage: {{ .Release.Service }}
+    app: "{{ .Chart.Name }}"
+    release: {{ .Release.Name | quote }}
+    heritage: {{ .Release.Service | quote }}
```
DandyDeveloper, line 10: Change these to the labels template. Technically `app` has been replaced by a newer label in Helm.

charts/opensearch/values.yaml
```diff
@@ -196,10 +196,17 @@ extraContainers: []
...
+ - name: volume-chmod
+   image: busybox
+   command: ['sh', '-c']
+   args:
+     - 'chown -R 1000:1000 /usr/share/opensearch'
+   securityContext:
+     runAsUser: 0
+   volumeMounts:
+     - mountPath: /usr/share/opensearch/data
+       name: opensearch-cluster-master
```
DandyDeveloper, line 209: If this is necessary, then it should be default to the statefulset.

DandyDeveloper (changes requested):

charts/opensearch/templates/securityconfig.yaml
alborotogarcia: Yes you're right! I didn't check twice!

charts/opensearch/values.yaml
alborotogarcia: done !

peterzhuamazon:
Need some reviews @DandyDeveloper @TheAlgo @smlx very much thanks :)

charts/opensearch/values.yaml
```diff
@@ -196,11 +196,6 @@ extraContainers: []
   #   image: busybox
   #   command: ['do', 'something']
 
-extraInitContainers: []
```
DandyDeveloper, line 199: We should keep this @alborotogarcia

DandyDeveloper (changes requested):
We need the `extraInitContainers` but empty:

```
extraInitContainers: []
```

charts/opensearch/values.yaml
alborotogarcia: I forgot to add it back nice catch @DandyDeveloper !

peterzhuamazon:
Hi @DandyDeveloper @TheAlgo Please review again seems there are changes from @alborotogarcia.
Thanks.

charts/opensearch/templates/statefulset.yaml
```diff
@@ -211,8 +211,20 @@ spec:
       {{- if .Values.hostAliases }}
       hostAliases: {{ toYaml .Values.hostAliases | nindent 8 }}
       {{- end }}
-      {{- if or (.Values.extraInitContainers) (.Values.keystore) }}
       initContainers:
+{{- if .Values.persistence.enabled }}
+      - name: fsgroup-volume
+        image: busybox
+        command: ['sh', '-c']
```
DandyDeveloper, line 218: Won't this break? `initContainers` will be written but if `persistence.enabled=`false`, `extraInitContainers=[]` AND `keystore.enabled=false` are missing. It's empty in the YAML.

DandyDeveloper (changes requested):
@alborotogarcia Sorry to be so pedantic here.

charts/opensearch/templates/statefulset.yaml
alborotogarcia: indeed @DandyDeveloper !  it should be init if any of them applies, now should be fixed

DandyDeveloper (approved):

TheAlgo (approved):
LGTM!!

sastorsl:
Regarding https://github.com/opensearch-project/helm-charts/pull/9/files#diff-08885706cd45ad445d2696cd7e39cf7d4b5e82cb51b0ee6bd86e6ee9f818e158R218

It would be nice if it was possible to override the image from values. For instance if one has a local image proxy with proper credentials setup.

PR #19, opened 2021-09-08
Use consistent k8s API semver comparison logic

smlx:
### Description
This is required to work around bugs in the version string returned by
kubernetes distros such as EKS and GKE, where they have invalid Semantic
Version strings. See https://github.com/helm/helm/issues/3810.
 
### Issues Resolved
Closes: #34 
 
### Check List

### DIFF
--- a/charts/opensearch-dashboards/templates/ingress.yaml
+++ b/charts/opensearch-dashboards/templates/ingress.yaml
@@ -29,7 +29,7 @@ spec:
     {{- end }}
   {{- end }}
   rules:
-  {{- if semverCompare ">=1.19" .Capabilities.KubeVersion.Version -}}
+  {{- if semverCompare ">=1.19-0" .Capabilities.KubeVersion.Version -}}
     {{- range .Values.ingress.hosts }}
     - host: {{ .host | quote }}
       http:


DandyDeveloper:
@smlx Talk me through how this solves that problem? Can't see the explanation in the linked issue.

smlx:
Okay, so this is a fun bug :smile: 

The major factor behind this issue is that EKS uses an invalid Semantic Version string in their Kubernetes distro: https://github.com/aws/eks-distro/issues/512

So when you try to use version comparison in a helm template, helm correctly recognises the version of Kubernetes as a _prerelease_ version, and since semver does not consider prerelease versions to satisfy version constraints by default, `semverCompare ">=1.19" .Capabilities.KubeVersion.Version` returns `false` for a Kubernetes version like `1.20-eks-123`.

The workaround for this problem is to specify a prerelease version string in the comparison. In this case, `1.19-0`. This causes `semverCompare` to also consider prerelease versions in the comparison.

The thing is that the ingress template _already_ uses this workaround elsewhere in the template to set the resource API version: https://github.com/opensearch-project/helm-charts/blob/main/charts/opensearch-dashboards/templates/ingress.yaml#L4

So what happens when you try to install on EKS is that the template sets Ingress API version `apiVersion: networking.k8s.io/v1`, but then uses the old `apiVersion: networking.k8s.io/v1beta1` object structure, which Kubernetes rightly rejects, resulting in the error message in my linked issue.

TheAlgo:
I see it being used elsewhere and looks like the best possible way to tackle the issue

TheAlgo (approved):

DandyDeveloper (approved):
@smlx Thank you for explaining. 

My only thing would be to mention this in the templating and highlight the issue in the template.

PR #21, opened 2021-09-08
Modify majorVersion fallback logic

smlx:
### Description
* Look in both .Values.imageTag and .Chart.AppVersion before falling
back to a default value.
* Use the built-in semver parsing function.
* Don't ignore the version for non-opensearch images.
 
### Issues Resolved
Closes: #32 
 
### Check List

### DIFF
--- a/charts/opensearch/templates/_helpers.tpl
+++ b/charts/opensearch/templates/_helpers.tpl
@@ -55,7 +55,7 @@ We truncate at 63 chars because some Kubernetes name fields are limited to this 
 {{- if .Values.majorVersion }}
   {{- .Values.majorVersion }}
 {{- else }}
-  {{- $version := semver (coalesce .Values.imageTag .Chart.AppVersion "7") }}
+  {{- $version := semver (coalesce .Values.imageTag .Chart.AppVersion "1") }}
   {{- $version.Major }}
 {{- end }}
 {{- end }}
--- a/charts/opensearch/templates/_helpers.tpl
+++ b/charts/opensearch/templates/_helpers.tpl
@@ -52,14 +52,10 @@ We truncate at 63 chars because some Kubernetes name fields are limited to this 
 {{- end -}}
 
 {{- define "opensearch.majorVersion" -}}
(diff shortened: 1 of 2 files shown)


DandyDeveloper (approved):

charts/opensearch/templates/_helpers.tpl
```diff
@@ -52,14 +52,10 @@ We truncate at 63 chars because some Kubernetes name fields are limited to this
...
-{{ $version }}
-  {{- else -}}
-7
-  {{- end -}}
-{{- end -}}
-{{- end -}}
+{{- if .Values.majorVersion }}
+  {{- .Values.majorVersion }}
+{{- else }}
+  {{- $version := semver (coalesce .Values.imageTag .Chart.AppVersion "7") }}
```
TheAlgo, line 58: Why we have made it `7`, shouldn't it be `1`?

smlx: The only reason that it is `7` is that I copied the fallback value that already existed (take a look at the diff to see this).

Should it be changed to `1`?

TheAlgo: Might be it was missed, but I feel we should have `1` as `7` is not yet released.  I don't think people will use this chart to install `elasticsearch` cluster.

@DandyDeveloper your views on this?

TheAlgo (changes requested):

charts/opensearch/templates/_helpers.tpl
TheAlgo: CC : @peterzhuamazon

smlx:
@TheAlgo I changed the fallback value from 7 to 1.

TheAlgo (approved):
LGTM!

charts/opensearch/templates/_helpers.tpl
peterzhuamazon: No need to include ODFE as it is in maintenance only mode.

Issue #40, opened 2021-09-10
[BUG][Opensearch] Difficult to share securityconfig between StatefulSets


smlx:
**Describe the bug**
It is difficult to share `securityConfig` values between Opensearch chart installs which are part of the same Opensearch cluster, when using the combined `securityConfig.config.data` value.

Currently the secret name must be defined in the `securityConfig.config.securityConfigSecret`, but it cannot be the same name for different chart installs in the same namespace.

**To Reproduce**
1. Define a shared `securityConfig.config.*` which is used by multiple chart installs which make up a single Opensearch cluster running in a single namespace.
2. See an error due to the Secret name conflict between chart installs.

**Expected behavior**
I'd expect the secret name to be auto-generated and not conflict between chart installs.

**Chart Name**
Opensearch

**Screenshots**
n/a

**Host/Environment (please complete the following information):**
 - Helm Version: v3.6.2
 - Kubernetes Version: 1.20.7
 
**Additional context**
n/a

PR #41, opened 2021-09-10
Use a per-install name for securityconfig secret

smlx:
### Description
Give the securityconfig secret an autogenerated unique name to
facilitate installing the chart multiple times in the same namespace.

This helps with the common case of sharing the securityconfig between
multiple instantiations of this chart to construct an Opensearch
cluster.
 
### Issues Resolved
Closes: #40 
 
### Check List


DandyDeveloper (commented):
The problem with this is the previous implementation allowed someone to pass in their own `template/secret.yaml` in their child chart. 

This actually becomes more restrictive.

We should support both. Maybe with a `{{ default (printf "%s-securityconfig" include "opensearch.uname" . ) .Values.securityConfig.config.securityConfigSecret }}` in the statefulset references to the volume?

smlx:
> The problem with this is the previous implementation allowed someone to pass in their own template/secret.yaml in their child chart.

Could you explain how this was supposed to work previously?

The volume bindings on the statefulset have conditionals like this, which is identical to the conditional on the creation of the `securityconfig` secret object:
```
{{- if and .Values.securityConfig.config.securityConfigSecret .Values.securityConfig.config.data }}
```
This means that they wouldn't have been able to supply an external secret name and have the statefulset use it with this mechanism.

I think you can still use an external secret by specifying `extraVolumes` + `extraVolumeMounts`?

DandyDeveloper:
@smlx I'm going to say this wasn't ported very well from the original. Those conditions are a bug, but I think your change still introduces more restrictions.

It'd be more logical to just remove the verbose (and unnecessary) conditions you've outlined.

smlx:
Ok, so what if we changed it to have this kind of logic:

* if you define anything under `data`, the chart will automatically create a secret and mount it (this is what the PR does now)
* if you define `securityConfigSecret`, the chart will assume this secret is created externally and try to mount it.
* it is an error to define both `data` and `securityConfigSecret`

Does that satisfy all the use cases? I'll obviously add some comments explaining how this works.

DandyDeveloper:
@smlx Perfect middle ground. Agreed.

TheAlgo:
> Ok, so what if we changed it to have this kind of logic:
> 
> * if you define anything under `data`, the chart will automatically create a secret and mount it (this is what the PR does now)
> * if you define `securityConfigSecret`, the chart will assume this secret is created externally and try to mount it.
> * it is an error to define both `data` and `securityConfigSecret`
> 
> Does that satisfy all the use cases? I'll obviously add some comments explaining how this works.

@smlx Can we add this explanation in the README for OpenSearch chart. It will be great to avoid any confusions for users.

peterzhuamazon:
Any progress on this PR @smlx ?

peterzhuamazon:
Sorry about my misclick on closing the PR, open again.
Hi @DandyDeveloper @TheAlgo can you please take a look and approve?
Thanks.

smlx:
I've rebased and bumped the version.

TheAlgo:
> > Ok, so what if we changed it to have this kind of logic:
> > 
> > * if you define anything under `data`, the chart will automatically create a secret and mount it (this is what the PR does now)
> > * if you define `securityConfigSecret`, the chart will assume this secret is created externally and try to mount it.
> > * it is an error to define both `data` and `securityConfigSecret`
> > 
> > Does that satisfy all the use cases? I'll obviously add some comments explaining how this works.
> 
> @smlx Can we add this explanation in the README for OpenSearch chart. It will be great to avoid any confusions for users.

@smlx Can we add it in the README as well?

smlx:
@TheAlgo ok, added a line to the README.

TheAlgo (commented):
@peterzhuamazon @mprimeaux @smlx @DandyDeveloper Should be make this `1.1.0`?

mprimeaux:
@peterzhuamazon Yes to `1.1.0` as this introduces a new non-breaking feature and therefore mandates a minor version number increment.

smlx:
OK, changed version to `1.1.0`.

DandyDeveloper:
@smlx I think that's the best way to address this.

PR #43, opened 2021-09-10
Copy of PR #45 from previous repo

hgoscenski-imanage:
https://github.com/opensearch-project/opensearch-devops/pull/45

Signed-off-by: Harrison Goscenski <harrison.goscenski@imanage.com>

### Description
[Describe what this change achieves]
 
### Issues Resolved
[List any issues this PR will resolve]
 
### Check List

### DIFF
--- a/charts/opensearch-dashboards/values.yaml
+++ b/charts/opensearch-dashboards/values.yaml
@@ -145,3 +145,15 @@ nodeSelector: {}
 tolerations: []
 
 affinity: {}
+
+probes:
+  readiness:
+    enabled: true
+    custom:
+      tcpSocket:
+        port: http
+      periodSeconds: 15
+  liveness:
+    enabled: true
+  startup:
+    enabled: true
--- a/charts/opensearch-dashboards/values.yaml
+++ b/charts/opensearch-dashboards/values.yaml
@@ -58,30 +58,24 @@ securityContext:
   runAsNonRoot: true
   runAsUser: 1000
 
-config:
-  opensearch_dashboards.yml:
-    server:
-      name: dashboards
-      host: 0.0.0.0
+config: {}
+  # Default OpenSearch Dashboards configuration from docker image of Dashboards
+  #  opensearch_dashboards.yml:
+  #   server:
+  #     name: dashboards
+  #     host: 0.0.0.0
 
-      ## Dashboards TLS Config
-      ssl:
-        enabled: false
-        key: /usr/share/opensearch-dashboards/certs/dashboards-key.pem
-        certificate: /usr/share/opensearch-dashboards/certs/dashboards-crt.pem
-    opensearch.ssl.certificateAuthorities: /usr/share/opensearch-dashboards/certs/dashboards-root-ca.pem
+  # Dashboards TLS Config (Ensure the cert files are present before enabling SSL
+      # ssl:
+      #   enabled: true
+      #   key: /usr/share/opensearch-dashboards/certs/dashboards-key.pem
+      #   certificate: /usr/share/opensearch-dashboards/certs/dashboards-crt.pem
 
-probes:
-  readiness:
-    enabled: true
-    custom:
-      tcpSocket:
-        port: http
-      periodSeconds: 15
-  liveness:
-    enabled: true
-  startup:
-    enabled: true
+    # determines how dashboards will verify certificates (needs to be none for default opensearch certificates to work)
+    # opensearch:
+    #   ssl:
+    #     certificateAuthorities: /usr/share/opensearch-dashboards/certs/dashboards-root-ca.pem
+    #     if utilizing custom CA certs for connection to opensearch, provide the CA here
 
 priorityClassName: ""
 
@@ -93,10 +87,10 @@ opensearchAccount:
 labels: {}
 
 hostAliases: []
-#- ip: "127.0.0.1"
-#  hostnames:
-#  - "foo.local"
-#  - "bar.local"
+# - ip: "127.0.0.1"
+#   hostnames:
+#   - "foo.local"
+#   - "bar.local"
 
 serverHost: "0.0.0.0"
 
--- a/charts/opensearch-dashboards/values.yaml
+++ b/charts/opensearch-dashboards/values.yaml
@@ -58,24 +58,30 @@ securityContext:
   runAsNonRoot: true
   runAsUser: 1000
 
-config: {}
-  ## Default OpenSearch Dashboards configuration from docker image of Dashboards
-  #  opensearch_dashboards.yml:
-  #   server:
-  #     name: dashboards
-  #     host: 0.0.0.0
+config:
+  opensearch_dashboards.yml:
+    server:
+      name: dashboards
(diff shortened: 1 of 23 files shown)


peterzhuamazon:
@DandyDeveloper @TheAlgo @smlx please take a look at this. Thanks.

charts/opensearch-dashboards/values.yaml
```diff
@@ -58,24 +58,30 @@ securityContext:
...
+    server:
+      name: dashboards
+      host: 0.0.0.0
+
+      ## Dashboards TLS Config
+      ssl:
+        enabled: false
+        key: /usr/share/opensearch-dashboards/certs/dashboards-key.pem
+        certificate: /usr/share/opensearch-dashboards/certs/dashboards-crt.pem
+    opensearch.ssl.certificateAuthorities: /usr/share/opensearch-dashboards/certs/dashboards-root-ca.pem
```
DandyDeveloper, line 72: Should this be commented?

charts/opensearch/templates/securityconfig.yaml
```diff
@@ -5,9 +5,9 @@ metadata:
   name: {{ .Values.securityConfig.config.securityConfigSecret }}
   namespace: {{ .Release.Namespace }}
   labels:
-    app: {{ .Chart.Name }}
-    release: {{ .Release.Name }}
-    heritage: {{ .Release.Service }}
+    app: { { .Chart.Name } }
+    release: { { .Release.Name } }
+    heritage: { { .Release.Service } }
```
DandyDeveloper, line 10: Weird looking code block here. Can you revert back?

charts/opensearch/templates/statefulset.yaml
```diff
@@ -137,70 +140,70 @@ spec:
...
-          secretName: {{ .Values.securityConfig.tenantsSecret }}
-      {{- end }}
+        {{- range .Values.secretMounts }}
+        - name: {{ .name }}
+          secret:
+            secretName: {{ .secretName }}
+            {{- if .defaultMode }}
+            defaultMode: {{ .defaultMode }}
+            {{- end }}
+        {{- end }}
```
DandyDeveloper, line 150: You've overwritten some master stuff in your merge. Please carefully re-add this.

charts/opensearch/templates/statefulset.yaml
```diff
@@ -224,31 +227,31 @@ spec:
...
 
-          {{ .Values.opensearchHome }}/bin/opensearch-keystore create
+          opensearch-keystore create
 
           for i in /tmp/keystoreSecrets/*/*; do
             key=$(basename $i)
             echo "Adding file $i to keystore key $key"
-            {{ .Values.opensearchHome }}/bin/opensearch-keystore add-file "$key" "$i"
+            opensearch-keystore add-file "$key" "$i"
           done
```
DandyDeveloper, line 236: See previous comments. You've overwridden in your merge.

DandyDeveloper (changes requested):
There's a number of unintentional changes based on your merge I suspect. 

I would recommend restarting from HEAD of master and adding those probes again.

peterzhuamazon:
Hi @hgoscenski-imanage could you address the issues commented by @DandyDeveloper ? Thanks

TheAlgo (commented):
@hgoscenski-imanage looks like you need to pull the latest changes from this repository. There are many stale changes in this PR. Thanks

charts/opensearch-dashboards/values.yaml
hgoscenski-imanage: Yes, have commented out.

charts/opensearch/templates/securityconfig.yaml
hgoscenski-imanage: My apologies, got lost in the transition between the repos and an overzealous formatter.

charts/opensearch/templates/statefulset.yaml
hgoscenski-imanage: I have removed the other range over secrets mounts as it has less options.

charts/opensearch/values.yaml
```diff
@@ -324,6 +324,37 @@ fullnameOverride: ""
...
+    interval: 10
+  # ensure the username and password used for READINESS_USERNAME and READINESS_PASSWORD has permissions to access /_cat/health otherwise the pod will never start
+  readiness:
+    enabled: false
+    interval: 10
+    delay: 30
+    # Default utilizes READINESS_USERNAME and READINESS_PASSWORD to query _cat/health for status
+    default: true
+    username: admin
+    password: admin
```
DandyDeveloper, line 344: Does the demo not expose a user that can do health endpoints checks? 

I know this is something users should actively resolve themselves. But I can see this scenario:

* User has their own security config / Uses defaults BUT changes admin
* Has their own parent chart without looking at full values
* Node never ready because they haven't updated this field with their new securityConfig credentials

Is there anyway we can template this based on an existing set of values? I had a quick look. I don't think there is.

DandyDeveloper: I guess the default accepts all states anyway, so maybe this is a non-issue.

DandyDeveloper (approved):
I just had 1 comment, this looks fine to me though. Thank you for the effort. Sorry its taken a long as it has to get this in.

peterzhuamazon:
Hi @smlx let me know if you can also take a look, as @TheAlgo is busy recently. Thanks.

charts/opensearch/values.yaml
hgoscenski-imanage: The default accepts all to avoid that specific issue, though they should definitely configure it correctly for their environment, if there is a larger meta-chart that defines good practices for a full deploy (master, client, data, dashboards) perhaps we set the defaults with a dedicated readiness probe user there?

hgoscenski-imanage: The other reason I think we ought to leave it as it is, is that is the users are not using a declarative setup and are instead using api calls to configure users they would then need to be able to start the first time without requiring a specific user and then redeploy with updated probes. I don't know how common that use case is, but I imagine it is at least some of the deployments.

smlx (changes requested):
I'm :-1: on this PR in its current state.

In general I think probes should be quick and simple. Authenticated `curl` requests are way too heavyweight.

In addition to that, I don't think liveness probes suit stateful, slow starting applications like Opensearch. If kubernetes starts killing pods all over the place when the cluster is unhealthy it's just going to make the situation worse.

So I'd really like to see simple connectivity readiness probes (e.g. is the port open) with the only option being `enabled`/`disabled`. Possibly also have an option to define fully `custom` probes via arbitrary yaml to suit other use-cases. But the way that the `probes` field is currently structured in `values.yaml` is over-parameterised. It would be far simpler as a freeform yaml field that is injected directly into the chart.

DandyDeveloper:
@smlx I already made these points as well, I was overruled.

TheAlgo:
Closing this PR for now, as there is no traction. Feel free to re-open

Issue #14, opened 2021-07-22
Publish helm charts


dblock:
Coming from opensearch-project/opensearch-devops#10 and opensearch-project/opensearch-devops#4 

consider https://github.com/helm/chart-releaser-action

TheAlgo:
@mprimeaux We have completed the migration. You should be good to contribute now. Thanks.

peterzhuamazon:
Hi @mprimeaux is there any progress on this task to make helm charts publish?
Thanks.

mprimeaux:
Hi @peterzhuamazon. I will issue a PR today. My apologies for the delay.

peterzhuamazon:
> Hi @peterzhuamazon. I will issue a PR today. My apologies for the delay.

np, thanks.

PR #46, opened 2021-09-14
Add support for Helm chart linting and releasing.

mprimeaux:
### Description
* Added support for the Helm chart testing action (linting and testing).
* Added support for the Helm chart releaser action.
* Fixed minor lint issues in Helm chart values files.
* Added GitHub Actions Workflow badges in README.
  
### Issues Resolved
[#14](https://github.com/opensearch-project/helm-charts/issues/14)
 
### Check List
- :white_check_mark: Commits are signed per the DCO using --signoff 

### DIFF
--- a/.github/workflows/lint-test.yaml
+++ b/.github/workflows/lint-test.yaml
@@ -1,0 +1,51 @@
+name: Lint and Test Charts
+
+on:
+  push:
+    branches-ignore:
+    - 'main'
+    - 'repository'
+    
+  pull_request:
+    branches:
+    - '**'
+
+jobs:
+  lint-test:
+    runs-on: ubuntu-20.04
+    steps:
+      - name: Checkout
+        uses: actions/checkout@v2
+        with:
+          fetch-depth: 0
+
+      - name: Install Helm
+        uses: azure/setup-helm@v1
+        with:
+          version: v3.4.1
+
+      - uses: actions/setup-python@v2
+        with:
+          python-version: '3.8'
+
+      - name: Install chart-testing
+        uses: helm/chart-testing-action@v2.1.0
+
+#      - name: Run chart-testing (list-changed)
+#        id: list-changed
+#        run: |
+#          changed=$(ct list-changed --config ct.yaml)
+#          if [[ -n "$changed" ]]; then
+#            echo "::set-output name=changed::true"
+#          fi
+
+      - name: Run chart-testing (lint)
+        run: ct lint --config ct.yaml
+
+#      - name: Create kind cluster
+#        uses: helm/kind-action@v1.2.0
+#        if: steps.list-changed.outputs.changed == 'true'
+
+#      # See https://github.com/helm/chart-testing/blob/main/doc/ct_install.md
+#      - name: Run chart-testing (install)
+#        run: ct install --config ct.yaml
--- a/.github/workflows/lint-test.yaml
+++ b/.github/workflows/lint-test.yaml
@@ -31,21 +31,21 @@ jobs:
       - name: Install chart-testing
         uses: helm/chart-testing-action@v2.1.0
 
-#      - name: Run chart-testing (list-changed)
-#        id: list-changed
-#        run: |
-#          changed=$(ct list-changed --config ct.yaml)
-#          if [[ -n "$changed" ]]; then
-#            echo "::set-output name=changed::true"
-#          fi
+      - name: Run chart-testing (list-changed)
+        id: list-changed
+        run: |
+          changed=$(ct list-changed --config ct.yaml)
+          if [[ -n "$changed" ]]; then
+            echo "::set-output name=changed::true"
+          fi
 
       - name: Run chart-testing (lint)
         run: ct lint --config ct.yaml
 
-#      - name: Create kind cluster
-#        uses: helm/kind-action@v1.2.0
-#        if: steps.list-changed.outputs.changed == 'true'
+      - name: Create kind cluster
+        uses: helm/kind-action@v1.2.0
+        if: steps.list-changed.outputs.changed == 'true'
 
-#      # See https://github.com/helm/chart-testing/blob/main/doc/ct_install.md
-#      - name: Run chart-testing (install)
-#        run: ct install --config ct.yaml
+      # See https://github.com/helm/chart-testing/blob/main/doc/ct_install.md
+      - name: Run chart-testing (install)
+        run: ct install --config ct.yaml --debug
--- a/.github/workflows/lint-test.yaml
+++ b/.github/workflows/lint-test.yaml
@@ -1,18 +1,10 @@
 name: Lint and Test Charts
 
-on:
-  push:
-    branches-ignore:
-    - 'main'
-    - 'repository'
-    
-  pull_request:
-    branches:
-    - '**'
-
+on: pull_request
+  
 jobs:
   lint-test:
-    runs-on: ubuntu-20.04
+    runs-on: ubuntu-latest
     steps:
       - name: Checkout
         uses: actions/checkout@v2
@@ -20,9 +12,9 @@ jobs:
           fetch-depth: 0
 
       - name: Install Helm
-        uses: azure/setup-helm@v1
+        uses: azure/setup-helm@v1.1
         with:
-          version: v3.4.1
+          version: v3.7.0
 
       - uses: actions/setup-python@v2
         with:
--- a/.github/workflows/release.yaml
+++ b/.github/workflows/release.yaml
@@ -7,7 +7,7 @@ on:
 
 jobs:
   release:
-    runs-on: ubuntu-20.04
+    runs-on: ubuntu-latest
     steps:
       - name: Checkout
         uses: actions/checkout@v2
@@ -22,14 +22,11 @@ jobs:
       - name: Install Helm
         uses: azure/setup-helm@v1
         with:
-          version: v3.4.1
+          version: v3.7.0
 
       # See https://github.com/helm/chart-releaser
       # See https://github.com/helm/chart-releaser-action
       - name: Run chart-releaser
         uses: helm/chart-releaser-action@v1.2.1
-        with:
-          config: cr.yaml
-          charts_dir: charts
         env:
           CR_TOKEN: "${{ secrets.GITHUB_TOKEN }}"
--- a/.github/workflows/release.yaml
+++ b/.github/workflows/release.yaml
@@ -1,0 +1,35 @@
+name: Release Charts
+
+on:
+  push:
+    branches:
+      - main
+
+jobs:
+  release:
+    runs-on: ubuntu-20.04
+    steps:
+      - name: Checkout
+        uses: actions/checkout@v2
+        with:
+          fetch-depth: 0
+
+      - name: Configure Git
+        run: |
+          git config user.name "$GITHUB_ACTOR"
+          git config user.email "$GITHUB_ACTOR@users.noreply.github.com"
+
+      - name: Install Helm
+        uses: azure/setup-helm@v1
+        with:
+          version: v3.4.1
+
+      # See https://github.com/helm/chart-releaser
+      # See https://github.com/helm/chart-releaser-action
+      - name: Run chart-releaser
+        uses: helm/chart-releaser-action@v1.2.1
+        with:
+          config: cr.yaml
+          charts_dir: charts
+        env:
+          CR_TOKEN: "${{ secrets.GITHUB_TOKEN }}"
--- a/.gitignore
+++ b/.gitignore
@@ -1,0 +1,3 @@
+# macOS
+.DS_Store
+
--- a/.gitignore
+++ b/.gitignore
@@ -1,3 +1,5 @@
 # macOS
 .DS_Store
 
+# Windows
+Thumbs.db
--- a/README.md
+++ b/README.md
@@ -16,6 +16,7 @@ A community repository for Helm Charts of OpenSearch Project.
 
 ```shell
 helm repo add opensearch https://opensearch-project.github.io/helm-charts/
+helm repo update
 helm search repo opensearch
 ```
 
--- a/README.md
+++ b/README.md
@@ -12,6 +12,16 @@
 
 A community repository for Helm Charts of OpenSearch Project.
 
+```shell
+helm repo add opensearch https://opensearch-project.github.io/helm-charts/
+helm search repo opensearch
+```
+
+## Status
+
+![Testing](https://github.com/opensearch-project/helm-charts/workflows/Lint%20and%20Test%20Charts/badge.svg)
+![Release](https://github.com/opensearch-project/helm-charts/workflows/Release%20Charts/badge.svg)
+
 ## Contributing
 
 See [developer guide](DEVELOPER_GUIDE.md) and [how to contribute to this project](CONTRIBUTING.md). 
--- a/README.md
+++ b/README.md
@@ -12,6 +12,8 @@
 
 A community repository for Helm Charts of OpenSearch Project.
 
+## Installation
+
 ```shell
 helm repo add opensearch https://opensearch-project.github.io/helm-charts/
 helm search repo opensearch
--- a/charts/opensearch-dashboards/values.yaml
+++ b/charts/opensearch-dashboards/values.yaml
@@ -59,13 +59,13 @@ securityContext:
   runAsUser: 1000
 
 config: {}
-  ## Default OpenSearch Dashboards configuration from docker image of Dashboards
+  # Default OpenSearch Dashboards configuration from docker image of Dashboards
   #  opensearch_dashboards.yml:
   #   server:
   #     name: dashboards
   #     host: 0.0.0.0
 
-  ## Dashboards TLS Config (Ensure the cert files are present before enabling SSL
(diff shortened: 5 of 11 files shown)


mprimeaux:
@peterzhuamazon @TheAlgo @DandyDeveloper @smlx  I set the GH pages branch as `repository` and not the default of `gh-pages`.  Please review and let me know if any changes are required. All comments and feedback are welcome.

.github/workflows/lint-test.yaml
```diff
@@ -0,0 +1,51 @@
...
+      - name: Install chart-testing
+        uses: helm/chart-testing-action@v2.1.0
+
+#      - name: Run chart-testing (list-changed)
+#        id: list-changed
+#        run: |
+#          changed=$(ct list-changed --config ct.yaml)
+#          if [[ -n "$changed" ]]; then
+#            echo "::set-output name=changed::true"
+#          fi
```
DandyDeveloper, line 40: I assume something is wrong with this because of helm 3? I found a similar problem on other repos.

.github/workflows/lint-test.yaml
```diff
@@ -0,0 +1,51 @@
...
+      - name: Run chart-testing (lint)
+        run: ct lint --config ct.yaml
+
+#      - name: Create kind cluster
+#        uses: helm/kind-action@v1.2.0
+#        if: steps.list-changed.outputs.changed == 'true'
+
+#      # See https://github.com/helm/chart-testing/blob/main/doc/ct_install.md
+#      - name: Run chart-testing (install)
+#        run: ct install --config ct.yaml
```
DandyDeveloper, line 51: This missing kind of defeats a core part of CI testing on the chart. 

What's the reason for this being removed?

.github/workflows/release.yaml
```diff
@@ -0,0 +1,35 @@
...
+
+      # See https://github.com/helm/chart-releaser
+      # See https://github.com/helm/chart-releaser-action
+      - name: Run chart-releaser
+        uses: helm/chart-releaser-action@v1.2.1
+        with:
+          config: cr.yaml
+          charts_dir: charts
+        env:
+          CR_TOKEN: "${{ secrets.GITHUB_TOKEN }}"
```
DandyDeveloper, line 35: @TheAlgo Someone with settings access will need to create this.

.gitignore
```diff
@@ -0,0 +1,3 @@
+# macOS
+.DS_Store
+
```
DandyDeveloper, line 3: Redundant? Maybe IDE specific?

charts/opensearch-dashboards/values.yaml
```diff
@@ -59,13 +59,13 @@ securityContext:
   runAsUser: 1000
 
 config: {}
-  ## Default OpenSearch Dashboards configuration from docker image of Dashboards
+  # Default OpenSearch Dashboards configuration from docker image of Dashboards
```
DandyDeveloper, line 62: Two ## better indicate a command. Is this a linting thing?

DandyDeveloper (commented):
Looks good for the most part, but I have some comments.

.github/workflows/lint-test.yaml
mprimeaux: The issue is not with helm 3 but with the fact that the GitHub runner only has 2 CPUs and 8 GB memory, which causes a failure with the default value of `replicas: 3`.  

The attached [log](https://github.com/opensearch-project/helm-charts/files/7171243/ct-install-log.txt) shows that `ct install` succeeds in testing both charts when using Minikube on my local machine with enough resources.

I can confirm that using `replicas: 1` in the values file for the `opensearch` chart within the GitHub runner succeeds.

The Kind GitHub action is fairly limited in what we can do with it and so I switched to the Minikube GitHub action. Setting `--cpus 2 --memory 6144mb`.  What fails is a replica count greater than 1 **but only within the GitHub runner**.

I do agree that we want to test and not only lint. I propose we (1) keep the `replicas: 3` in the values file and (2) override it to `1` in the GitHub actions workflow, which does provide some level of testing. 

What are your thoughts?

.github/workflows/lint-test.yaml
mprimeaux: I included a reply to this comment above.

.github/workflows/release.yaml
mprimeaux: There is no need to create this as it's [provided by GitHub](https://docs.github.com/en/actions/reference/authentication-in-a-workflow).

.gitignore
mprimeaux: This is for macOS and unrelated to an IDE. macOS creates this file when you visit a folder in Finder. It is needed to avoid adding it inadvertently.

charts/opensearch-dashboards/values.yaml
mprimeaux: Yes, it's a linting thing.

.gitignore
peterzhuamazon: I want to raise the bar and see if we can add all these:
```
Thumbs.db
.DS_Store

.idea/
*.iml

__pycache__
.pytest_cache

.coverage
coverage.xml

.gradle/
.settings/
out.txt
```

peterzhuamazon: This is the standard one we use in opensearch-build repo.

mprimeaux: Agreed. I'll add the above.

mprimeaux: If we really want to 'go wild' then there is [this](https://github.com/github/gitignore).

.github/workflows/lint-test.yaml
mprimeaux: After a bit of testing various options, we have a few options:

**Option 1:** Leave `replicas: 3` and use an external GitHub runner. 

I don't really like this option.

**Option 2:** Set `replicas: 1` and include a note indicating the value should not be set to a value of less than 3 for production operational environments.

This is also a less than ideal option because, by default, this results in a single copy of the index. So folks would almost always need to override this value. But this will allow tests to pass.

The reason why this option is 'less than ideal' is because OpenSearch sets the index replica count to 1 and there's a rule that disallows the index replica to be deployed to the same pod (a.k.a. OpenSearch node).

**Option 3:** Leave `replicas: 3` and 'fix-up' the `charts/opensearch/values.yaml` file **before** running chart testing.

I think this is our best option as it tests the chart and keeps minimum operating values 'as is'. In the end, we aren't testing OpenSearch but are testing the chart.

mprimeaux: For the curious, [here](https://github.com/mprimeaux/helm-charts/runs/3613235595?check_suite_focus=true) is a successful run with linting and testing.  I simply used `sed` in a custom action to modify the values file.

charts/opensearch-dashboards/values.yaml
DandyDeveloper: Space wasn't the concern here, you removed ## -> #. I'm wondering if that part is intentional?

.gitignore
DandyDeveloper: I didn't know MacOS added those default, sorry!

.github/workflows/release.yaml
DandyDeveloper: Weird, I swear back in the day I had to make this. Thank you for the link!

charts/opensearch-dashboards/values.yaml
mprimeaux: The `##` needed to be `# ` to avoid the yaml lint error.

.github/workflows/lint-test.yaml
peterzhuamazon: We have tried option 1 back in ODFE and it is painful to maintain a provision script, bootstrap to GitHub code, and the cleanup code. Sometimes the cleanup is not working and we have to manually delete the runner in GitHub.

I like Option 3 and agree with your comments. We are testing if charts work.

DandyDeveloper: By fix-up, we mean override with something like `--set replicas=2`? 

I don't see it in the step: https://github.com/mprimeaux/helm-charts/runs/3613235595?check_suite_focus=true#step:10:30 

Maybe I'm misunderstanding. Sorry.

smlx: you can override the defaults in CI by creating a `ci/ci-values.yaml` file inside the chart.

> Charts may have multiple custom values files matching the glob pattern '*-values.yaml' in a directory named 'ci' in the root of the chart's directory. The chart is installed and tested for each of these files. If no custom values file is present, the chart is installed and tested with defaults.

From https://github.com/helm/chart-testing/blob/main/doc/ct_install.md

smlx: I'd recommend putting values in here to scale CI to be as small and fast as possible: no persistence, single replica etc.

smlx: > The Kind GitHub action is fairly limited in what we can do with it and so I switched to the Minikube GitHub action.

I'm quite familiar with `kind` and the github action. What issues did you run into here?

The nice thing about `kind` is the speed compared to `minikube`.

mprimeaux: @smlx Thanks for pointing out the `ci` folder. I didn't see that option when reviewing the `ct install` command. I do think this is the best option.

Regarding the use of Minikube over Kind, I should have clarified that I switched to Minikube in an attempt to use all resources (cpu and memory) available on the GitHub runner by adding the following step:

```
# See https://github.com/marketplace/actions/setup-minikube-kubernetes-cluster
- name: Create minikube cluster
  uses: manusa/actions-setup-minikube@v2.4.2
  if: steps.list-changed.outputs.changed == 'true'
  with:
    minikube version: 'v1.23.0'
    kubernetes version: 'v1.21.4'
    driver: docker
    github token: ${{ secrets.GITHUB_TOKEN }}
    start args: '--addons metrics-server --cpus 2 --memory 6144mb'
```

This isn't needed anymore since `replicas: 1` will be set in the CI-specific values file.

mprimeaux: Hi @DandyDeveloper 

> By fix-up, we mean override with something like `--set replicas=2`?
> 
> I don't see it in the step: https://github.com/mprimeaux/helm-charts/runs/3613235595?check_suite_focus=true#step:10:30
> 
> Maybe I'm misunderstanding. Sorry.

I had created the following step before running `ct install`:

```
- name: Fixup values file
  run: |
    sed -i 's/replicas: 3/replicas: 1/g' charts/opensearch/values.yaml
```

Based on what @smlx pointed out, this step of course isn't needed in lieu of the CI-specific values file.

mprimeaux: Hi @smlx 

> The nice thing about `kind` is the speed compared to `minikube`.

FWIW, the duration difference between Minikube and Kind for their respective steps was marginal:

* Kind: [1 min 49 seconds](https://github.com/mprimeaux/helm-charts/runs/3621760978?check_suite_focus=true)
* Minikube: [1 min 14 seconds](https://github.com/mprimeaux/helm-charts/runs/3613235595?check_suite_focus=true)

Nevertheless, as suggested, I reverted to using Kind.

mprimeaux:
@smlx @DandyDeveloper @TheAlgo @peterzhuamazon I believe I've addressed everyone's feedback and sincerely appreciate the time you all took to review. 

Please let me know if there are any other concerns or feedback.

.github/workflows/lint-test.yaml
TheAlgo: I also kinda agree on using `kind` only. Thanks @mprimeaux for this.

TheAlgo (approved):
Changes looks good to me.

.github/workflows/lint-test.yaml
peterzhuamazon: Kind +1

DandyDeveloper (approved):
Thank you for clarifying everything. Great start.

TheAlgo:
@smlx @peterzhuamazon Can you review the latest changes as per the feedback. We can push this PR after that

.github/workflows/lint-test.yaml
```diff
@@ -0,0 +1,51 @@
...
+    steps:
+      - name: Checkout
+        uses: actions/checkout@v2
+        with:
+          fetch-depth: 0
+
+      - name: Install Helm
+        uses: azure/setup-helm@v1
+        with:
+          version: v3.4.1
```
smlx, line 25: This is pretty old, could we bump it to the latest version?

.github/workflows/lint-test.yaml
```diff
@@ -0,0 +1,51 @@
...
+
+on:
+  push:
+    branches-ignore:
+    - 'main'
+    - 'repository'
+    
+  pull_request:
+    branches:
+    - '**'
```
smlx, line 11: I'm not sure what the workflow on this repository is, but if `main` is protected from pushes and PRs are required then this can be just
```
on: pull_request
```

.github/workflows/lint-test.yaml
```diff
@@ -0,0 +1,51 @@
...
+
+jobs:
+  lint-test:
+    runs-on: ubuntu-20.04
```
smlx, line 15: minor nit, but `ubuntu-latest` means this doesn't need to be bumped every few years :slightly_smiling_face:

.gitignore
```diff
@@ -0,0 +1,5 @@
+# macOS
```
smlx, line 1: This file doesn't belong in the repository, instead it should go in your system git configuration.

https://git-scm.com/docs/gitignore

cr.yaml
```diff
@@ -0,0 +1,4 @@
+# See https://github.com/helm/chart-releaser-action
```
smlx, line 1: I think this file is unnecessary since these values are automatically detected by the action: https://github.com/helm/chart-releaser-action/blob/main/action.yml#L21-L22

And the `pages-branch` is `gh-pages` by default, which matches the github pages default branch which is used as the helm repository. Any reason why it is not `gh-pages` here?

ct.yaml
```diff
@@ -0,0 +1,11 @@
...
+# See https://github.com/helm/chart-testing#configuration
+remote: origin
+target-branch: main
+chart-dirs:
+  - charts
+chart-repos: []
+helm-extra-args: --timeout 800s
+check-version-increment: false
+validate-maintainers: false
+validate-yaml: true
```
smlx, line 11: IMO all the checks should be enabled. Any other opinions?

smlx (changes requested):
A few comments and questions but a great start overall.

.github/workflows/lint-test.yaml
mprimeaux: Sure.

.github/workflows/lint-test.yaml
mprimeaux: Sure. I'll modify it.

I had pinned it due to the lag when `latest` still resulted in Ubuntu 18.04 LTS even though Ubuntu 20.04 LTS was released. It was a bit frustrating since `latest` didn't actually mean latest.

.gitignore
mprimeaux: Thanks. I wasn't aware of that subtle difference.

cr.yaml
mprimeaux: It's a personal preference to convey more semantic meaning rather than the general name of `gh-pages`. Happy to change since there is enough context given this repository is solely dedicated to Helm charts.

ct.yaml
mprimeaux: I agree.

.github/workflows/lint-test.yaml
mprimeaux: I wasn't sure of the workflow, either. The intention is to lint and test everywhere except the `main` and `gh-pages` (formerly `repository`) branches. 

I'll assume `on: pull_request` is correct as per your suggestion.

mprimeaux:
@smlx I believe your comments have been addressed. Please let me know.

peterzhuamazon:
I have no objections, will let @smlx to confirm the changes from @mprimeaux before merge.
Thanks everyone.

.github/workflows/lint-test.yaml
peterzhuamazon: I personally like to use specific number as latest can change and some code does not work the same way between OS versions sometimes. This one tho, I am not too worried to use latest as it is just a lint test.

README.md
```diff
@@ -12,6 +12,16 @@
 
 A community repository for Helm Charts of OpenSearch Project.
 
+```shell
```
TheAlgo, line 15: Sorry to bring it now, but can we not add a title for installation here,

mprimeaux: Done.

mprimeaux:
@smlx Please let me know if any other concerns need to be addressed. We are hoping to leverage this PR in our release this week. Your support is appreciated.

mprimeaux:
I am just checking in to see if there are any other outstanding changes being requested.  Thanks much.

peterzhuamazon:
Thanks @mprimeaux I see you have addressed all findings, I am also waiting for @smlx to final confirm.
We will check back again tomorrow, and will merge this if no more comments.


Thanks.

peterzhuamazon:
Not seeing any more comments will approve now.
Thanks.

Issue #67, opened 2021-09-27
[Enhancement][opensearch] make initcontainer image of persistency configurable


bitnik:
**Is your feature request related to a problem? Please describe.**

Right now when persistent volume for opensearch is enabled, chart uses a hard-coded image for an initcontainer:

https://github.com/opensearch-project/helm-charts/blob/dd0c692a9ca1dc97baa5931b67187bdfc333824a/charts/opensearch/templates/statefulset.yaml#L216-L219

Since we can use images only from our own registry, we can't deploy this chart directly, but we have to customise it.

**Describe the solution you'd like**

Make it configurable via values.yaml

```yaml
persistence:
  enabled: true
  image: busybox
  imageTag: latest
```

and update necessary templates.

-------

If you aggree on the change, I can create a PR to add it.

sastorsl:
Will also affect where image requests are proxied internally to ensure proper licensing.
A big :+1:

PR #70, opened 2021-09-28
make initcontainer image of persistency configurable

bitnik:
### Description

This PR makes  initcontainer image of persistency configurable.
 
### Issues Resolved

https://github.com/opensearch-project/helm-charts/issues/67
 
### Check List


peterzhuamazon:
I like this approach but not sure if it is really needed. 
It is busybox after all and I currently dont see an issue with hardcoding.
What is your take on this @DandyDeveloper @smlx ?

Thanks.

DandyDeveloper (commented):
I agree with this change but I'm confused on why its under persistence? I also wonder if its necessary? Because additional initContainers will be added by users and will need their full spec anyway.

bitnik:
> It is busybox after all and I currently dont see an issue with hardcoding.

As stated in the https://github.com/opensearch-project/helm-charts/issues/67 we use a private image registry and we can only use images from there, so we need to modify the image name and prepend the registry URL. Is there any other way to change t he image registry URL, maybe we are missing it?

> I agree with this change but I'm confused on why its under persistence?

I added under persistence, because right now it is the only place missing. I can also update the PR and add a config `imageRegistry`, and if it is set, we can prepend it to any image. What do you think?

bitnik:
@DandyDeveloper and @peterzhuamazon I updated thr PR and added a global `dockerRegistry` config to be used for all images. Could you review please?

peterzhuamazon:
I like this and has triggered the link/test action.
@DandyDeveloper @TheAlgo please check.

Thanks.

DandyDeveloper (commented):
@bitnik Isn't this typically provided in the `image.repository` anyway? 

If this is just a blanket value you want to globally apply, we should consider this going under a `global.xxx` position in the yaml. 

Can you make that change?

bitnik:
> @bitnik We should consider this going under a global.xxx position in the yaml.
Can you make that change?

Done.

> Also, what do we think about imageRepositoryPrefix over dockerRegistry?

I prefer `dockerRegistry` or `imageRegistry`, because the values we put in would be usually URLs of hosted private registries.

DandyDeveloper (approved):
All good with this assuming others are happy.

TheAlgo:
> @bitnik We should consider this going under a `global.xxx` position in the yaml.
> 
> Can you make that change?
> 
> Also, what do we think about `imageRepositoryPrefix` over `dockerRegistry`?
> 
> @peterzhuamazon @TheAlgo

@DandyDeveloper I also prefer `dockerRegistry`. Anything specific you have in mind w.r.t `imageRepositoryPrefix` ?

bitnik:
Btw should I do a version bump?

peterzhuamazon:
> Btw should I do a version bump?

Yes, please do a version bump. Thanks.

bitnik:
Version bump is done, and also updated the changelog.

peterzhuamazon:
> Version bump is done, and also updated the changelog.

Thanks I have no more comments. If @TheAlgo approve we can merge this.

bitnik:
@TheAlgo could you please review? The new release would make things really easier for us.

PR #71, opened 2021-09-28
Added basic support for plugins on nodes

### LINKED ISSUE #50: [Enhancement][Opensearch] Support OpenSearch Prometheus Exporter
Ported from: https://github.com/opensearch-project/opensearch-devops/issues/68

Add support for the exporter. 

Best done by mounting an initContainer that will write the plugins. This can be a bit of a pain for persistence between containers, but I will look into using the existing PVC volume to install the plugins against. 

N/R 

**Additional context**
N/R

DandyDeveloper:
### Description
Adds the ability to define plugins. This will enable a basic bash script on the opensearch container to install the plugins.

Unfortunately, we cannot manage this in an initContainer because the directory in question by default has plugins AND would be overwritten if we defined it as a new volume. 
 
### Issues Resolved
- fixes #50 
 
### Check List


DandyDeveloper:
@mprimeaux Side question for this, do we want to be adding additional CI values to enable the explicit functionality that we'd want to pick up problem on?

mprimeaux:
> @mprimeaux Side question for this, do we want to be adding additional CI values to enable the explicit functionality that we'd want to pick up problem on?
> 
> In this case, I have it disabled by default, but ideally we want the CI picking up problems if this is enabled.

@DandyDeveloper I do think it is best to reason through what is the set of CI values files are needed to maximize the 'code coverage' of the chart. Might be worth time decomposing the [MECE](https://en.wikipedia.org/wiki/MECE_principle) model of the areas to test and then creating a values file for each.

DandyDeveloper:
@mprimeaux We can define the model, or maybe simplify things into a basic list of "scenarios" and then try and build out the numerous values that encompasses all those scenarios? 

Or, would is there a better way to do this? 

I don't know how best to translate an MECE model into a helm chart series to "tests" (Without spending more time then should be spent making it).

DandyDeveloper:
@mprimeaux Lets move this to an issue and get another PR out for this. 

I think I can add some basic attributes to the CI for this to run a reasonable test on my specific addition here.

mprimeaux:
@DandyDeveloper Agreed.

DandyDeveloper:
@mprimeaux @TheAlgo I need some thumbs.

mprimeaux:
@DandyDeveloper LGTM

mprimeaux:
@DandyDeveloper when we finally merge PR #75 then all changes other than chart-related will not cause linting and testing to trigger.

charts/opensearch/values.yaml
```diff
@@ -367,3 +367,9 @@ fsGroup: ""
 ## Also see: https://kubernetes.io/docs/tasks/administer-cluster/sysctl-cluster/
 sysctl:
   enabled: false
+
+## Enable to add 3rd Party / Custom plugins not offered in the default OpenSearch image.
+plugins:
+  enabled: false
+  installList: []
+  # - https://github.com/aparo/opensearch-prometheus-exporter/releases/download/1.0.0/prometheus-exporter-1.0.0.zip
```
TheAlgo, line 375: IMO we should not give this as an example and rather keep it empty and add instructions for installation in the README. Many of them can think it as an official recommendation as well.  Rest of the changes looks good to me.

DandyDeveloper: @TheAlgo I thought this might be the case. The problem is it'd be nice to have an official example, but I can't find any. haha

DandyDeveloper:
@TheAlgo Think we're good for your thumbs now.

DandyDeveloper:
@smlx @peterzhuamazon Need one of you to approve as well per the repo requirements.

smlx:
I don't have write access, so I don't think my approval will help :slightly_smiling_face:

TheAlgo:
> @TheAlgo Think we're good for your thumbs now.

Yes we are good now.

TheAlgo (changes requested):
One small thing. the chart version needs to be bumped right? 😄

DandyDeveloper:
@TheAlgo

TheAlgo (approved):

TheAlgo:
@peterzhuamazon Have a look when free. Thanks

peterzhuamazon:
I dont have any objections on this. Looks good to me.
However, many PRs right now all have 1.0.5 version we needs to find a way to get this work.

DandyDeveloper:
@peterzhuamazon We need to be a little "faster" at getting thing in I think. Alternatively, maybe we could make a feature branch that'll get merged upstream every X period? 

I.E. feature-1.5 -> people branch from this and PRs get merged upstream into that feature branch? 

@smlx @mprimeaux Just floating ideas, because Peter is right, there's a lot of contributions floating. The current process only works with less frequent user contributions. Otherwise it can be overwhelming and users may forget their PRs after we've asked for another update.

mprimeaux:
@smlx @DandyDeveloper Let's compartmentalize into the two related tracks of discussion.

**Track 1: Branching Strategy:**  I think the predominant question to answer here is do the maintainers want to introduce long-live release branches? My vote is an absolute "yes".

**Track 2: Version Automation:** I think [Conventional Commits](https://www.conventionalcommits.org/en/v1.0.0/) is a good thing.  How the implementation in GitHub Actions workflow is implemented, in part, *may* depend on the decisions (output) of Track 1.  For example, will PRs to `main` be treated any differently than PRs to the long-lived release branches. If the answer is simply 'no' then that's fine but we can at least reach a consensus so it's covered. The `main` branch would be treated simply as the next version or edge branch.

DandyDeveloper:
We really need a Slack or Zulip channel for these discussions :)

mprimeaux:
💯

TheAlgo:
> Meta: This PR thread is going way off track - maybe it should be moved to a discussion?
> 
> IMO long lived release branches have many issues.
> 
> The first is versioning. If you have multiple long lived release branches how do you accurately follow SemVer without conflicting versions?
> 
> The second is complexity for contributors. Imagine I'm a contributor who submits a PR with a fix. Which branch do I submit it to? And how long until I can use the fix in a released chart? With the current single branch the answers are `main` and "as soon as it is merged" thanks to release automation. With long lived release branches the answers are much less clear. This is a huge disincentive for contributors.
> 
> Third is complexity for maintainers. This repository already struggles to review and merge PRs in a timely manner. Is AWS going to employ several more people to manage maintenance across half a dozen independent release branches?
> 
> Fourth is tooling: you're going to be fighting it and adding lots of complexity to CI to implement long-lived release branches.
> 
> > I would personally vote for Track1, we can create a dev branch to let people keep contributing without version issues. Then maintainer can create PRs with bumped version from dev merge to main in much slower pace.
> 
> Why would you want to go from fast fully-automated releases to slow, manual releases? If a PR is merged, shouldn't that code be released immediately? And if the concern is that new code is not adequately tested, then the answer is to improve CI, not to slow down the release process.

I am in favour of this approach completely. CI improvement is a good suggestion. Let's not stop agile development.

PR #102, opened 2021-10-21
Add option to disable initContainer chown update

### LINKED ISSUE #91: [Enhancement][opensearch] Make volume permissions init container optional with persistence enabled
We were using helm chart version 1.0.0 from the old repository and trying to migrate to the newer versions from this one we found out that a new init container to set the file system permissions was added when persistence is enabled. We are using in some of our k8s instances EFS with dynamic provisioning which doesn't allow to set the permissions on the file system causing the init container to fail and thus the node cannot start.

This wasn't an issue on the previous version without the init container as the EFS CSI driver takes care of setting the permissions so the chown is not needed in that case.

It should be possible to disable the volume permissions init container when persistence is enabled. Maybe a new property can be added to the values file to enable/disable the init container instead of just checking if persistence is enabled.

We are still using version 1.0.0 of the chart as it doesn't have that init container and it works fine for now. Going forward the solution would probably be maintaining a fork of the chart but it would be preferable to have the toggle option in the values file.

**Additional context**
No additional context.

DandyDeveloper:
### Description
Adds a basic option to disable the `fsGroup-volume` chown permissions update initContainer. Per request in #91 
 
### Issues Resolved
- fixes #91 
 
### Check List

### DIFF
--- a/charts/opensearch/CHANGELOG.md
+++ b/charts/opensearch/CHANGELOG.md
@@ -19,6 +19,16 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Support to disable the initContainer `fsgroup-volume` for chown updates.
 
 ---
+## [1.2.2]
+### Added
+### Changed
+### Deprecated
+### Removed
+### Fixed
+- [Issue #105](https://github.com/opensearch-project/helm-charts/issues/105) OpenSearch chart fails when RBAC is enabled.
+### Security
+
+---
 ## [1.2.1]
 ### Added
 ### Changed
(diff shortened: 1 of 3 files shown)


charts/opensearch/CHANGELOG.md
```diff
@@ -13,6 +13,11 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ### Fixed
 ### Security
 
+---
+## [1.2.1]
```
mprimeaux, line 17: Should this be updated to `1.2.2` to match Chart version?

peterzhuamazon:
I have no more comments other than @TheAlgo and @mprimeaux .

PR #172, opened 2021-12-13
Add Readiness probe to OpenSearch Chart

### LINKED ISSUE #170: [Enhancement][OpenSearch] Add readiness Probe to the chart
Have a [readiness probe](https://kubernetes.io/docs/tasks/configure-pod-container/configure-liveness-readiness-startup-probes/) for OpenSearch charts

Add readiness probe in Stateful set template which can check for the cluster health status to be green. It will have some timeout which can be controlled via `values.yaml`.

N/A

**Additional context**
We already have the option in `values.yaml` but it is not getting used at the moment.

TheAlgo:
### Description
Configure readiness probe in the chart. The configs were present in `values.yaml` but it was not getting used. 
### Issues Resolved
#170 
 
### Check List
- [X] Commits are signed per the DCO using --signoff 


peterzhuamazon:
Need more eyes @smlx @mprimeaux if you have time please take a look.

Thanks.

DandyDeveloper:
I'm going to need to take a bit to consume this one properly. It's definitely an interesting readiness probe...

mprimeaux:
Same. I'll also dedicate time first thing in the morning to reason through the solution domain.

TheAlgo:
@mprimeaux @DandyDeveloper It is still in progress. I would be happy if I can get an early feedback

bkomraz1:
@TheAlgo 

Here is fix for readiness_probe

1. By default ELASTIC_USERNAME/ELASTIC_USERNAME should be admin/admin as it defined in default internal_users.yml
2. https protocol should be used for readiness_probe

```
Index: charts/opensearch/templates/statefulset.yaml
IDEA additional info:
Subsystem: com.intellij.openapi.diff.impl.patch.CharsetEP
<+>UTF-8
===================================================================
diff --git a/charts/opensearch/templates/statefulset.yaml b/charts/opensearch/templates/statefulset.yaml
--- a/charts/opensearch/templates/statefulset.yaml	(revision ba44d9fcfd2d6106113ae9f3bd33932609c2f242)
+++ b/charts/opensearch/templates/statefulset.yaml	(revision 1d897e00be1271e34499264cbf13ad1083db0c0e)
@@ -311,8 +311,8 @@
                   if [ "$args" != "" ]; then
                     set -- "$@" $args
                   fi
-                  if [ -n "${USERNAME}" ] && [ -n "${PASSWORD}" ]; then
-                    set -- "$@" -u "${USERNAME}:${PASSWORD}"
+                  if [ -n "${ELASTIC_USERNAME:-admin}" ] && [ -n "${ELASTIC_PASSWORD:-admin}" ]; then
+                    set -- "$@" -u "${ELASTIC_USERNAME}:${ELASTIC_PASSWORD}"
                   fi
                   curl --output /dev/null -k "$@" "{{ .Values.protocol }}://127.0.0.1:{{ .Values.httpPort }}${path}"
                 }
@@ -341,7 +341,7 @@
                     exit 1
                   fi
                 fi
-{{ toYaml .Values.readinessProbe | indent 10 }}
+          {{ toYaml .Values.readinessProbe | nindent 10 }}
         ports:
         - name: http
           containerPort: {{ .Values.httpPort }}
Index: charts/opensearch/values.yaml
IDEA additional info:
Subsystem: com.intellij.openapi.diff.impl.patch.CharsetEP
<+>UTF-8
===================================================================
diff --git a/charts/opensearch/values.yaml b/charts/opensearch/values.yaml
--- a/charts/opensearch/values.yaml	(revision ba44d9fcfd2d6106113ae9f3bd33932609c2f242)
+++ b/charts/opensearch/values.yaml	(revision 1d897e00be1271e34499264cbf13ad1083db0c0e)
@@ -250,7 +250,7 @@
 # If you experience slow pod startups you probably want to set this to `false`.
 enableServiceLinks: true

-protocol: http
+protocol: https
 httpPort: 9200
 transportPort: 9300
```

peterzhuamazon:
@TheAlgo any more activities on this PR?
Thanks.

TheAlgo:
> @TheAlgo any more activities on this PR? Thanks.

Yes the code changes are done. I will need to work through the CI issues. Did not get time. Will try to wrap this up and it will be ready for review early next week.

DandyDeveloper:
@TheAlgo You can probably also triple check the official ES way of doing this in their charts. It's very similar to this.

TheAlgo:
> @TheAlgo You can probably also triple check the official ES way of doing this in their charts. It's very similar to this.

Thanks @DandyDeveloper. I will take a look how they are addressing this and make this PR ready.

TheAlgo:
@DandyDeveloper @smlx @peterzhuamazon Please have a look when free. Thanks

TheAlgo:
> @DandyDeveloper @smlx @peterzhuamazon Please have a look when free. Thanks

+@mprimeaux

mprimeaux (approved):
LGTM

peterzhuamazon:
@DandyDeveloper @smlx if you guys can take a final look it would be great.
Thanks.

TheAlgo:
@DandyDeveloper Can you take out some time to go through this PR. Thanks in advance

DandyDeveloper:
@TheAlgo This is on my radar. Sorry, been all over the place lately. Need to give this some proper time and rollouts locally to test.

TheAlgo:
> @TheAlgo This is on my radar. Sorry, been all over the place lately. Need to give this some proper time and rollouts locally to test.

@DandyDeveloper Any luck on this?

DandyDeveloper:
@TheAlgo No. I'm still being pulled in my personal time to a number of things. I may commit some time to this tomorrow since I know you're waiting. I'm sorry to keep you waiting.

peterzhuamazon:
@prudhvigodithi would you please help checking this PR as well?
Thanks.

charts/opensearch/values.yaml
```diff
@@ -100,9 +100,11 @@ config:
...
 # This will be appended to the current 'env:' key. You can use any of the kubernetes env
 # syntax here
-extraEnvs: []
-#  - name: MY_ENVIRONMENT_VAR
-#    value: the_value_goes_here
+extraEnvs:
+  - name: PASSWORD
+    value: admin
+  - name: USERNAME
+    value: admin
```
prudhvigodithi, line 108: Hey @TheAlgo can we have the password field blank and allow user to pass in, if still left blank we can generate a random password in secret, we can use helm randAlphaNum(randAlphaNum 20 | b64enc ) string function and add it to a secret as follows
```
          - name: PASSWORD
            valueFrom:
              secretKeyRef:
                name: {{ template "opensearch.uname" . }}-credentials
```
This way the default password(admin:admin) is not added in charts by default and user can still override it passing in values.yaml file.

prudhvigodithi: Hey, just check this issue https://github.com/opensearch-project/security/issues/1576
adding directly as admin could miss from users end the action to change it, so which is why I have passed my above comment, but i'm ok as long as we add a comment above `- name: PASSWORD` saying that its recommended to change the password.

TheAlgo: That makes sense to add some more docstring and a line item in the docs as well. I can do that once @DandyDeveloper also completed reviewing this.

TheAlgo:
@smlx Do you mind taking a look at this?

charts/opensearch/values.yaml
TheAlgo: @prudhvigodithi I have made the changes

peterzhuamazon:
@DandyDeveloper @prudhvigodithi please review again as this PR has been dangling for too long 😅 
Thanks.

prudhvigodithi:
@TheAlgo I have tested the chart installation and looks good, it took 4m27s for the pod to be up after adding Readiness, just curious where do I check the output of this Readiness probe script?

TheAlgo:
> @TheAlgo I have tested the chart installation and looks good, it took 4m27s for the pod to be up after adding Readiness, just curious where do I check the output of this Readiness probe script?

For failures it comes up when we describe the pods. Also in the opensearch logs you should be able to see the health check logs

prudhvigodithi:
> For failures it comes up when we describe the pods. Also in the opensearch logs you should be able to see the health check logs

Thanks.

prudhvigodithi (approved):

smlx:
I see significant problems with using an authenticated readiness probe:

1. What happens when I change the password and/or admin user. Do I then also have to update `extraEnvs` to avoid the pods never becoming ready?
2. The readiness probe should not use the `admin` user. It should use a user with minimal permissions instead.

Is there any way that the readiness probe can be unauthenticated? That would remove a huge amount of complexity.

TheAlgo:
> I see significant problems with using an authenticated readiness probe:
> 
> 1. What happens when I change the password and/or admin user. Do I then also have to update `extraEnvs` to avoid the pods never becoming ready?
> 2. The readiness probe should not use the `admin` user. It should use a user with minimal permissions instead.
> 
> Is there any way that the readiness probe can be unauthenticated? That would remove a huge amount of complexity.
> 
> In addition, the cluster does _not_ have to be `green` to be usable. Many administrative actions can cause the cluster to become temporarily `yellow` and still be perfectly usable.
> 
> I don't think the readiness probe should be inspecting the state of the application (e.g. green status). If the HTTP API endpoint is available (e.g. returns a response with HTTP status <500) then IMO Kubernetes should consider it to be ready and should be sending it traffic.


For the first point, the username and password can be changed by the user of the chart. It should just have the permission to check the cluster health of the cluster. Yes the `extraEnvs` need to be changed to make the readiness probe work

Also there is an option using which you can edit the `clusterHealthCheckParams` . So the use case of `yellow` can be handled by the chart user there.


@smlx How does this look now?

smlx:
> Also there is an option using which you can edit the clusterHealthCheckParams . So the use case of yellow can be handled by the chart user there.

Imagine the scenario where this change is applied and then the cluster goes `red`. How will you fix it? Kubernetes will refuse to send any traffic to any Opensearch pods because they aren't `Ready`, so you can't access the HTTP API via the Kubernetes `Service`.

smlx:
> For the first point, the username and password can be changed by the user of the chart. It should just have the permission to check the cluster health of the cluster. Yes the extraEnvs need to be changed to make the readiness probe work.

Imagine another scenario where I update the credentials of the `admin` user via the HTTP API. The cluster will immediately go unavailable because Kubernetes will consider it `NotReady`. How would you recover from that case?

prudhvigodithi:
Hey @smlx and @TheAlgo, Please check my following analysis

Any username and password configuration should be happening from helm chart, its expected when OpenSearch is installed via helm, manual changes should not be done via HTTP calls, unless these changes are not supported with helm.

From [cluster health](https://opensearch.org/docs/latest/opensearch/rest-api/cluster-health/), each call should be authenticated, so the username and password should be passed else we get `Unauthorized`. So to check health of a cluster we need some authentication.

Also how about we add just this 
`curl -k -u 'admin:admin' https://localhost:9200 --silent --fail`
```
          readinessProbe:
            exec:
              command:
              - /bin/bash
              - -c
              - curl -k -u ${USER}:${PASS} --silent --fail
                https://localhost:9200
            failureThreshold: 3
            initialDelaySeconds: 30
            periodSeconds: 30
            successThreshold: 1
            timeoutSeconds: 1
```
This will just query the cluster and check if OpenSearch is running or not, I think should be more simple and with less complexity

smlx:
> From [cluster health](https://opensearch.org/docs/latest/opensearch/rest-api/cluster-health/), each call should be authenticated, so the username and password should be passed else we get Unauthorized. So to check health of a cluster we need some authentication.

Cluster health indicates the state of the entire cluster. Kubernetes pod readiness indicates only the state of the individual node. These are two significantly different concepts. A pod is healthy when it is able to service requests. It is not relevant whether the response to these requests shows that the cluster is healthy or unhealthy.

Elasticsearch introducing a dedicated TCP port listener for readiness probes which only becomes available once the node is in a `Ready` state. There is some good discussion on the relevant issue and PR around why they prefer lightweight, unauthenticated health probes:

* https://github.com/elastic/elasticsearch/issues/81168
* https://github.com/elastic/elasticsearch/pull/84375

While the Elasticsearch solution is ideal, in the absence of such a feature in Opensearch I think Kubernetes pod readiness should simply rely on the HTTP API port being open.

prudhvigodithi:
> Cluster health indicates the state of the entire cluster. Kubernetes pod readiness indicates only the state of the individual node. These are two significantly different concepts. A pod is healthy when it is able to service requests. It is not relevant whether the response to these requests shows that the cluster is healthy or unhealthy.

The node is part of the cluster, so when a node can accept a traffic when cluster is healthy, as mentioned we dont have any endpoint to check the health of an individual node as part of cluster, hence the ideal way is to check cluster health which sums up if the nodes are ready to serve the traffic or not.

Those links are helpful, since there is no dedicated port, we might need to go with @TheAlgo solution or just add `curl -k -u 'admin:admin' https://localhost:9200 --silent --fail`

smlx:
> Any username and password configuration should be happening from helm chart, its expected when OpenSearch is installed via helm, manual changes should not be done via HTTP calls, unless these changes are not supported with helm.

If this is the case, it is not documented. And regardless, changes are not possible via helm because changes to config files won't be loaded into the `.opendistro_security` index unless you exec into a pod and [manually run `securityadmin.sh`](https://opensearch.org/docs/latest/security-plugin/configuration/security-admin/#a-word-of-caution). Any change to the config files via helm will not update the running configuration automatically, even with a pod restart.

> Those links are helpful, since there is no dedicated port, we might need to go with @TheAlgo solution or just add `curl -k -u 'admin:admin' https://localhost:9200 --silent --fail`

If you want to encode that logic you can just use a native Kubernetes [HTTP probe](https://kubernetes.io/docs/tasks/configure-pod-container/configure-liveness-readiness-startup-probes/#define-a-liveness-http-request) instead of `curl`, because "Any code greater than or equal to 200 and less than 400 indicates success. Any other code indicates failure."

However since Opensearch nodes can return `503` in some "normal" cases such as when discovering master nodes after a restart I wonder if all the nodes restarting at once (thus losing quorum) could lead to the cluster never becoming ready? There are probably other similar pathological cases. This problem is also raised in the linked Elasticsearch issue in [this comment](https://github.com/elastic/elasticsearch/issues/81168#issuecomment-1039474371).

TheAlgo:
All interesting discussions. Thanks @smlx and @prudhvigodithi for all your points. I went through the PR in elasticsearch and I also agree that is is a near perfect solution not necessarily the best w.r.t to security but as long as they can disable it , I don't see an issue. @prudhvigodithi also gave an interesting endpoint that we can use but again it also needs authentication. It paves the way to eliminate `/health` endpoint and other red/yellow cluster issues but has issues of inconsistent security metadata. 

Authenticated probes has its own pros and cons. Considering OpenSearch is a complex distributed system and running it on k8s it more complex it might become tricky for users to come out of some weird issues. Saw some intermittent github issues because of the readiness probe which was implemented by elasticsearch operator as well. 

Now what I propose is a native TCP readiness probe something like this : 
```
readinessProbe:
  failureThreshold: 10
  tcpSocket:
     port: 9200
  initialDelaySeconds: 10

```
@smlx @prudhvigodithi Thoughts?

smlx:
Maybe adding a startup probe would be better than the `initialDelaySeconds`? Something like this?
```
startupProbe:
  tcpSocket:
    port: 9200
  failureThreshold: 30
  periodSeconds: 10
readinessProbe:
  tcpSocket:
    port: 9200
```

TheAlgo:
@prudhvigodithi Final opinions on this?

prudhvigodithi:
Hey @TheAlgo and @smlx I'm good with startupProbe, as @smlx suggested we can add something like this 
as soon as the startupProbe succeeds this probe is deactivated and the Liveness probe starts to be executed.
```
startupProbe:
  tcpSocket:
    port: 9200
  initialDelaySeconds: 5
  periodSeconds: 10
  timeoutSeconds: 3
  failureThreshold: 30
readinessProbe:
  tcpSocket:
    port: 9200
  periodSeconds: 5
  timeoutSeconds: 3
  failureThreshold: 3
 ```
 Lets do a full round of test, so that this is not blocking the startup times for the cluster.

**Comment by DandyDeveloper** [comment]
Sorry, I've been completely ignorant of this topic the past few weeks. If you do need my opinion, let me know. Looks like a lot of good discussion has happened since I last looked.

**Comment by TheAlgo** [comment]
> Sorry, I've been completely ignorant of this topic the past few weeks. If you do need my opinion, let me know. Looks like a lot of good discussion has happened since I last looked.

Feel free to jump in @DandyDeveloper

**Comment by peterzhuamazon** [comment]
This has been dangling for way too long.
What would be the path forward @DandyDeveloper @TheAlgo @prudhvigodithi 

Also bare in mind we recently cut 1.x branch so please consider if this PR requires to be pushed on both branch.

https://github.com/opensearch-project/helm-charts#version-and-branching

Thanks.

**Comment by TheAlgo** [comment]
> This has been dangling for way too long. What would be the path forward @DandyDeveloper @TheAlgo @prudhvigodithi @smlx
> 
> Also bare in mind we recently cut 1.x branch so please consider if this PR requires to be pushed on both branch.
> 
> https://github.com/opensearch-project/helm-charts#version-and-branching
> 
> Thanks.

@peterzhuamazon Apologies did not get time to address some latest minor comments and testing. Will try to wrap it up by this week. Also yes it needs to go in both the branches.

**Comment by TheAlgo** [comment]
@prudhvigodithi @peterzhuamazon I have updated the PR. Please have a look when free. Thanks

File: charts/opensearch/ci/ci-values.yaml [hunk 2]
```diff
@@ -320,12 +322,21 @@ terminationGracePeriod: 120
...
+    port: 9200
+  initialDelaySeconds: 5
+  periodSeconds: 10
+  timeoutSeconds: 3
+  failureThreshold: 30
+
 readinessProbe:
   failureThreshold: 3
-  initialDelaySeconds: 10
+  initialDelaySeconds: 900
```
smlx, line 336: Is 15 minutes too long to fire in CI? Maybe there should also be a test using the default values?

charts/opensearch/ci/ci-rbac-enabled-values.yaml
```diff
@@ -318,12 +320,21 @@ terminationGracePeriod: 120
 
 sysctlVmMaxMapCount: 262144
 
+startupProbe:
+  enabled: false
```
smlx, line 324: This value doesn't seem to be enabled anywhere in the CI values? Also I think having `enabled` under this key will break the template because it just inserts the whole contents of `startupProbe`, expecting it to be valid probe YAML.

charts/opensearch/values.yaml
```diff
@@ -323,12 +323,22 @@ terminationGracePeriod: 120
...
+  initialDelaySeconds: 5
+  periodSeconds: 10
+  timeoutSeconds: 3
+  failureThreshold: 30
 readinessProbe:
+  tcpSocket:
+    port: 9200
+  periodSeconds: 5
+  timeoutSeconds: 3
   failureThreshold: 3
```
smlx, line 341: This is the default value so can be removed.

charts/opensearch/values.yaml
```diff
@@ -323,12 +323,22 @@ terminationGracePeriod: 120
...
+  tcpSocket:
+    port: 9200
+  initialDelaySeconds: 5
+  periodSeconds: 10
+  timeoutSeconds: 3
+  failureThreshold: 30
 readinessProbe:
+  tcpSocket:
+    port: 9200
+  periodSeconds: 5
```
smlx, line 339: Any particular reason why this has been halved from the default of 10 seconds?

smlx (changes requested):
Some minor issues but looks pretty good.

charts/opensearch/ci/ci-ingress-class-name-values.yaml
```diff
@@ -318,12 +320,21 @@ terminationGracePeriod: 120
...
+    port: 9200
+  initialDelaySeconds: 5
+  periodSeconds: 10
+  timeoutSeconds: 3
+  failureThreshold: 30
+
 readinessProbe:
   failureThreshold: 3
-  initialDelaySeconds: 10
+  initialDelaySeconds: 900
```
DandyDeveloper, line 334: 15 minutes is a long time. CI doesn't have an implicit timeout does it?

DandyDeveloper (commented):
I think I'm coming into this quite late but one thing I don't see is how authentication is managed through the endpoints? Is the health endpoint now unauthenticated on the Opensearch API?

charts/opensearch/ci/ci-ingress-class-name-values.yaml
TheAlgo: > I think I'm coming into this quite late but one thing I don't see is how authentication is managed through the endpoints? Is the health endpoint now unauthenticated on the Opensearch API?

So @DandyDeveloper previously we were trying to have an authenticated probe but that has a lot of cons which was overweighing the pros. You can follow the discussion [here](https://github.com/opensearch-project/helm-charts/pull/172#issuecomment-1114815558)

charts/opensearch/values.yaml
TheAlgo: Just added it for good experience. Good to mention this explicitly so that the novice users know what is happening. Opinions?

charts/opensearch/ci/ci-values.yaml
TheAlgo: I have seen it take this much time for a 50 node cluster which I was trying to bring up. We can bring down the value, it should not be an issue

charts/opensearch/ci/ci-rbac-enabled-values.yaml
TheAlgo: The reason I added the flag was because of the api support for lower version cluster. Read [more](https://github.com/helm/helm/issues/7731#issuecomment-594993686). Somehow CI was not supporting the probe. . I will add a version check instead and remove the enabled key itself. How does that go in?

charts/opensearch/values.yaml
TheAlgo: There is no solid reason, just some hit and trial experiments.

sdwerwed:
Can we move this PR forward @prudhvigodithi @peterzhuamazon ? `kubect rollout restart`  is causing downtimes without this feature

charts/opensearch/values.yaml
peterzhuamazon: I am ok with it staying there, even thought it is defaults seems like good references.

peterzhuamazon:
@TheAlgo I see @smlx @DandyDeveloper have given some great comments.
Let me know if we can move forward with either changes or we can merge it as it be?
If these are not hard blockers we can always try to change later.

Thanks.

smlx:
> Can we move this PR forward @prudhvigodithi @peterzhuamazon ? `kubect rollout restart` is causing downtimes without this feature

This feature won't help with a rolling restart causing the cluster to go Yellow or Red, because that's related to overall cluster state. This feature is just probing individual node readiness.

charts/opensearch/ci/ci-rbac-enabled-values.yaml
TheAlgo: @smlx Are we good to close this comment?

charts/opensearch/ci/ci-values.yaml
TheAlgo: @smlx Are we good to close this comment?

charts/opensearch/ci/ci-ingress-class-name-values.yaml
TheAlgo: @DandyDeveloper Are we good to close this comment?

charts/opensearch/values.yaml
```diff
@@ -323,12 +323,21 @@ terminationGracePeriod: 120
 
 sysctlVmMaxMapCount: 262144
 
+clusterHealthCheckParams: "wait_for_status=green&timeout=1s"
```
prudhvigodithi, line 326: Hey @TheAlgo, I dont think so we need this as well `clusterHealthCheckParams:`, its not being used anywhere in templates, please correct me if i'm wrong.

TheAlgo: Yes @prudhvigodithi we can remove this now since we are not using authenticated probe anymore

charts/opensearch/ci/ci-ingress-class-name-values.yaml
```diff
@@ -318,12 +320,21 @@ terminationGracePeriod: 120
 
 sysctlVmMaxMapCount: 262144
 
+startupProbe:
+  enabled: false
```
prudhvigodithi, line 324: Hey @TheAlgo, thanks for the fix, do we still need this `enabled` flag ?

TheAlgo: Not required, I missed this CI value , removing it

TheAlgo:
> @TheAlgo I see @smlx @DandyDeveloper have given some great comments. Let me know if we can move forward with either changes or we can merge it as it be? If these are not hard blockers we can always try to change later.
> 
> Thanks.

I have addressed a critical comment, rest of them were more on clarifications and adjustments which can be picked later once people start using it on an adhoc basis

prudhvigodithi (approved):

TheAlgo:
@smlx Can you take a look at this PR when free?

charts/opensearch/ci/ci-values.yaml
smlx: Yep!

charts/opensearch/README.md
```diff
@@ -104,7 +104,10 @@ helm uninstall my-release
...
 | `volumeClaimTemplate`              | Configuration for the [volumeClaimTemplate for StatefulSets][]. You will want to adjust the storage (default `30Gi` ) and the `storageClassName` if you are using a different storage class                                                               | see [values.yaml][]                             |
-| `extraObjects`                     | Array of extra K8s manifests to deploy                                                                                                                                                                                                                    | list `[]`                                       |
+| `extraObjects`                     | Array of extra K8s manifests to deploy                                                                                                                                                                                                                    | list `[]`                                       |                                   |
```
smlx, line 107: This line has an extra `|` at the end, and the one below is missing the last `|`. Also the `|` are not aligned with the lines below.

charts/opensearch/ci/ci-values.yaml
```diff
@@ -320,12 +322,20 @@ terminationGracePeriod: 120
 
 sysctlVmMaxMapCount: 262144
 
+startupProbe:
```
smlx, line 325: I think all these overrides should be removed so that the actual chart defaults are tested in CI. i.e. remove the `startupProbe` and `readinessProbe` blocks from this file entirely.

What do you think?

charts/opensearch/values.yaml
```diff
@@ -397,6 +404,7 @@ networkPolicy:
   ## Example for default configuration to access transport port:
   ## opensearch-master-transport-client: "true"
 
+
```
smlx, line 407: Is this new empty line intentional?

charts/opensearch/values.yaml
```diff
@@ -250,7 +250,7 @@ podManagementPolicy: "Parallel"
 # If you experience slow pod startups you probably want to set this to `false`.
 enableServiceLinks: true
 
-protocol: http
+protocol: https
```
smlx, line 253: Why does the protocol here have to change?

smlx (changes requested):
A few more questions but it's getting there. :slightly_smiling_face:

smlx:
I also wonder if you could squash some of the commits down - 54 commits seems pretty big!

charts/opensearch/README.md
TheAlgo: Good nit catch. I will fix this indentation

charts/opensearch/values.yaml
TheAlgo: > I also wonder if you could squash some of the commits down - 54 commits seems pretty big!

We will squash and merge @smlx

TheAlgo:
> I also wonder if you could squash some of the commits down - 54 commits seems pretty big!

It shows my hardwork. Lol. Anyways @smlx we will squash and merge when we merge to main

charts/opensearch/values.yaml
TheAlgo: I changed the default value for some testing, can revert it back, no rationale

charts/opensearch/values.yaml
TheAlgo: Not really 🙂

charts/opensearch/ci/ci-values.yaml
TheAlgo: In one of the CI runs we can remove, that should be fine

charts/opensearch/README.md
TheAlgo: Not sure. the preview of the change looks clean to me.

charts/opensearch/values.yaml
TheAlgo: Removed

charts/opensearch/ci/ci-values.yaml
TheAlgo: We can do that

TheAlgo:
@smlx I have addressed your comments. Please review. Thanks for your continued support

TheAlgo:
@peterzhuamazon Adding the PR to your radar as well

peterzhuamazon (approved):
Thanks @TheAlgo for 6 months of back and forth hard work 😄 
Please also backport this to 1.x if needed.

Thanks.

TheAlgo:
> 

Sure thing

PR #200, opened 2022-01-22
securityConfig mounts

bkomraz1:
### Description
Accurate mount of securityConfig files
 
### Issues Resolved
default files are disappear from securityConfig folder after mount of custom security file
[#207]
 
### Check List


peterzhuamazon:
@DandyDeveloper @TheAlgo please take a look and review this PR.
Thanks.

TheAlgo (commented):
@bkomraz1 Can you take a look at the CI failures?

charts/opensearch/templates/statefulset.yaml
```diff
@@ -57,6 +57,9 @@ spec:
         {{- if .Values.config }}
         configchecksum: {{ include (print .Template.BasePath "/configmap.yaml") . | sha256sum | trunc 63 }}
         {{- end }}
+        {{- if .Values.securityConfig.config.data }}
+        securityconfigchecksum: {{ include (print .Template.BasePath "/securityconfig.yaml") . | sha256sum | trunc 63 }}
+        {{- end }}
```
DandyDeveloper, line 62: There are other conditions where this basePath may be used. It'd be nice to have an `or` here to add in when the existing secret mount has been presented.

charts/opensearch/templates/statefulset.yaml
```diff
@@ -366,8 +369,11 @@ spec:
           subPath: tenants.yml
         {{- end }}
         {{- if .Values.securityConfig.config.data  }}
-        - mountPath: {{ .Values.securityConfig.path }}
-          name: security-config-complete
+        {{- range $key, $_ := .Values.securityConfig.config.data  }}
+        - mountPath: {{ $.Values.securityConfig.path }}/{{ $key }}
+          name: security-config
```
DandyDeveloper, line 374: Doesn't this need to be `name: security-config-{{ $key }} or some kind of unique identifier? Otherwise this breaks since you are using subPath and the secret can have multiple keys.

DandyDeveloper (changes requested):
@bkomraz1 Have some concerns on the mounting, I could be wrong. I haven't testing this locally.

charts/opensearch/templates/statefulset.yaml
bkomraz1: You can use the same volume for mounting separate files (and not entirely volume content) by using subPath. And you can do it multiple times.

charts/opensearch/templates/statefulset.yaml
bkomraz1: It is only in the case when securityConfig.config.data is defined. This condition is from here:

```
{{- if .Values.securityConfig.config.data -}}
apiVersion: v1
kind: Secret
metadata:
  name: {{ include "opensearch.uname" . }}-securityconfig
```

So if chart is using exist secret, securityConfig.config.data can't be used. As result "securityconfig" secret will not be generated and securityconfigchecksum annotation will not be calculated.

charts/opensearch/templates/statefulset.yaml
DandyDeveloper: @bkomraz1 You can. Stupid me. Thank you.

DandyDeveloper (approved):

peterzhuamazon:
@TheAlgo please approve and I will fix the changelogs and versions.
Thanks.

bkomraz1:
@peterzhuamazon I resolved conflicts with latest main

peterzhuamazon:
> @peterzhuamazon I resolved conflicts with latest main

Thanks, will wait for @TheAlgo approval on this.

peterzhuamazon:
@TheAlgo can you give a approval if @bkomraz1 satisfied all the changes?
Thanks.

smlx (changes requested):
This is a breaking change and would require a major version bump because it implements a change in behaviour: the security config you define is now _merged_ with the upstream demo security config, instead of replacing it.

IMO this PR implements surprising behaviour because with it any changes made to the demo security config in the upstream image will now impact my security config unless I have explicitly overridden it locally. The demo security config from the upstream image also has some stuff that should never make it into production e.g.
https://github.com/opensearch-project/security/blob/e0be1ebdfc8577fdbed7b636fc3bec03e90ca3a9/securityconfig/internal_users.yml#L11-L18

A better fix might be to either define a default set of security configs in `values.yaml` for `securityConfig.config.data."config.yml"` etc. Or alternatively to better document the way this currently behaves so that it doesn't surprise people.

bkomraz1:
@smlx @peterzhuamazon 
I added support for securityConfig.config.dataComplete flag that is enabled by default. In this case usage of securityConfig.config.data will causes to disabling all default securityConfig files (it is current behaviour).
If value is set to false content of security folder will not disappear but specified files will be replaced.

peterzhuamazon:
cc: @TheAlgo

TheAlgo:
@DandyDeveloper @peterzhuamazon Do we want to bump the chart to `1.8.0` here?

peterzhuamazon:
> @DandyDeveloper @peterzhuamazon Do we want to bump the chart to `1.8.0` here?

I will bump after you guys approve the code part.

TheAlgo (approved):
LGTM!. Safe to merge after minor chart bump

peterzhuamazon:
Thanks @bkomraz1 for the contribution here!
It takes a long time to merge this.
Appreciate your help and support all the way.

Issue #205, opened 2022-01-24
[CI] Need automate version bump PR to be opened once OpenSearch/Dashboards release new versions


peterzhuamazon:
As of now, we are relying on human efforts to bump OS/OSD versions once there is a new release.
This could delay the release on these platforms and could add to human errors.
We need a github action to automatically open a PR after detecting the new releases.

TheAlgo:
I am still of the opinion to have release branches for charts 😄 

The maintainers can branch out from `main` and have releases enabled only on branches apart from `main`. 

I

viceice:
[renovate](https://github.com/renovatebot/renovate) can so this in an automated way 🤗

Issue #173, opened 2021-12-13
[Enhancement][opensearch] Release notes


sastorsl:
**Is your feature request related to a problem? Please describe.**
Release notes include the name of the component and version number, but no release notes.
This means you have to go hunting for what was released / changed / breaking.

Example https://github.com/opensearch-project/helm-charts/releases/tag/opensearch-1.5.4

**Describe the solution you'd like**
Automatically include release notes for all releases, both opensearch, opensearch-dashboards and future charts.

Example: https://docs.github.com/en/repositories/releasing-projects-on-github/automatically-generated-release-notes

**Describe alternatives you've considered**
N/A

**Additional context**
There are a few examples out there, i.e. https://github.com/helm/chart-releaser/releases

peterzhuamazon:
Hi @sastorsl so you mean we should include the release notes in the release page, not just in the CHANGELOG.md file?
If so, do you have any solutions that can contribute as a PR?
Thanks.

sastorsl:
I don't have a solution ready on hand no, but I guess there are at least some examples around.
I.e. the latest release from Prometheus https://github.com/prometheus/prometheus/releases/tag/v2.32.0

sastorsl:
Could for instance commit messages since the last merge be included in the release notes?

sastorsl:
@peterzhuamazon would more examples help?
I'm not personally familiar with how to achieve this, but there are examples in upstream of the _outcome_, both on what is included in the releases page on github, but also what is included in the RSS feeds.

It is really helpful when following the RSS-feeds since we are constantly reviewing tens / hundres of packages.

peterzhuamazon:
@DandyDeveloper @smlx thoughts?

sastorsl:
Did the team get a chance to look into this?

sastorsl:
@peterzhuamazon as a minimum, could the Release Note point to the merged PR?

Ref https://github.com/opensearch-project/helm-charts/releases/tag/opensearch-dashboards-1.4.0
It would help if one could click directly on the link to the PR.

TheAlgo:
PR link is something we can get from the commit in the release but the release note can be a bit more verbose. There are many interesting examples in the link which @smlx shared

> See here: https://github.com/helm/chart-releaser/issues/118

PR #257, opened 2022-05-03
Add 2.x appVersion as optional version for helm charts

peterzhuamazon:
Signed-off-by: Peter Zhu <zhujiaxi@amazon.com>

### Description
Add 2.x appVersion as optional version for helm charts
 
### Issues Resolved
https://github.com/opensearch-project/opensearch-build/issues/1624
 
### Check List

### DIFF
--- a/charts/opensearch-dashboards/CHANGELOG.md
+++ b/charts/opensearch-dashboards/CHANGELOG.md
@@ -13,6 +13,15 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ### Fixed
 ### Security
 ---
+## [1.5.1]
+### Added
+- Add optional appVersion for 2.x OpenSearch-Dashboards release
+### Changed
+### Deprecated
+### Removed
+### Fixed
+### Security
+---
 ## [1.4.1]
 ### Added
 - Update `values.yaml` example to use the correct format for configs
@@ -194,7 +203,8 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ### Fixed
 ### Security
 
-[Unreleased]: https://github.com/opensearch-project/helm-charts/compare/opensearch-dashboards-1.4.1...HEAD
+[Unreleased]: https://github.com/opensearch-project/helm-charts/compare/opensearch-dashboards-1.5.1...HEAD
+[1.5.1]: https://github.com/opensearch-project/helm-charts/compare/opensearch-1.4.1...opensearch-1.5.1
 [1.4.1]: https://github.com/opensearch-project/helm-charts/compare/opensearch-1.4.0...opensearch-1.4.1
 [1.4.0]: https://github.com/opensearch-project/helm-charts/compare/opensearch-1.3.3...opensearch-1.4.0
 [1.3.3]: https://github.com/opensearch-project/helm-charts/compare/opensearch-1.3.2...opensearch-1.3.3
--- a/charts/opensearch-dashboards/Chart.yaml
+++ b/charts/opensearch-dashboards/Chart.yaml
@@ -22,10 +22,10 @@ version: 1.5.1
 # follow Semantic Versioning. They should reflect the version the application is using.
 # It is recommended to use it with quotes.
 
-# 1.x Version Latest
+# 1.x Latest Version
 appVersion: "1.3.1"
 
-# 2.x Version Latest
+# 2.x Latest Version
 # appVersion: "2.0.0-rc1"
 
 maintainers:
--- a/charts/opensearch-dashboards/Chart.yaml
+++ b/charts/opensearch-dashboards/Chart.yaml
@@ -15,13 +15,18 @@ type: application
 # This is the chart version. This version number should be incremented each time you make changes
 # to the chart and its templates, including the app version.
 # Versions are expected to follow Semantic Versioning (https://semver.org/)
-version: 1.4.1
+version: 1.5.1
 
 # This is the version number of the application being deployed. This version number should be
 # incremented each time you make changes to the application. Versions are not expected to
 # follow Semantic Versioning. They should reflect the version the application is using.
 # It is recommended to use it with quotes.
+
+# 1.x Version Latest
 appVersion: "1.3.1"
+
+# 2.x Version Latest
+#appVersion: "2.0.0-rc1"
 
 maintainers:
   - name: DandyDeveloper
--- a/charts/opensearch-dashboards/Chart.yaml
+++ b/charts/opensearch-dashboards/Chart.yaml
@@ -26,7 +26,7 @@ version: 1.5.1
 appVersion: "1.3.1"
 
 # 2.x Version Latest
-#appVersion: "2.0.0-rc1"
+# appVersion: "2.0.0-rc1"
 
 maintainers:
   - name: DandyDeveloper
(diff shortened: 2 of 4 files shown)


peterzhuamazon:
@DandyDeveloper @TheAlgo @prudhvigodithi we need to think about how to support multiple versions.
As of now I dont think adding version branch is better than just allowing user to switch version themselves.
But I welcome ideas on this one.

The thing is, our release action will keep pushing new chart versions from main.
This means if we change appVersion to 2.0.0-rc1 now two days later 1.3.2 release will change it back to 1.x.
Which is not ideal at all.

Thanks.

prudhvigodithi:
Hey @peterzhuamazon, thats true.

- Main branch should always point to latest release (Consider main as a development branch).
- We already have tags to use a specific version.

So lets say if we have 2.0.0-rc1 release, main should have 2.0.0-rc1, we need to increment the version to `2.x` and cut a tag with this, user should refer to tag rather than main branch. Same goes with 1.3.2, main should have latest release 1.3.2 and cut a tag with `1.x`, so that users can use this tag who still uses 1.3.x version.

peterzhuamazon:
Per @bbarani we are not switching the current running version to rc1 and will keep as `1.3.1` until ga.
This PR is still useful to allow users to know there is an option.
Thanks.

prudhvigodithi:
> Per @bbarani we are not switching the current running version to rc1 and will keep as `1.3.1` until ga.
> This PR is still useful to allow users to know there is an option.
> Thanks.

Then we might need to cut a tag manually as opensearch-2.0.0-rc1 (that has `appVersion: "2.0.0-rc1"`), so that users can directly use this tag for 2.0.0-rc1, thoughts @bbarani and @peterzhuamazon ?

TheAlgo:
My 2 cents:

1.  Adding a documented line item will not make much sense to me , this is something as default version. This can ideally point to any OpenSearch version. We can have proper documentation regarding how to upgrade versions via helm charts and maybe maintain a compatibility matrix of the helm chart version and the OpenSearch version if any. Ideally main should point to latest release.
2.  Also we should invest some effort in better release notes #173 

3. On the release I am still of the opinion to not release so many chart versions and rather have release branches for charts which we can release periodically and non periodically in case of urgencies. Check comments in #205

bbarani (approved):

peterzhuamazon:
Hi,

Thanks @TheAlgo @prudhvigodithi on your suggestions here.
We still have some time before 2.0.0 GA comes.
This PR can stay here for a bit and I will probably put back to draft.

@smlx @mprimeaux what is your take on this one?

Thanks.

smlx:
What is the support policy for Opensearch 1.x upstream? Will they be doing further 1.x releases? Will it be getting security patches for some period of time? If so you might need a 1.x branch.

peterzhuamazon:
> What is the support policy for Opensearch 1.x upstream? Will they be doing further 1.x releases? Will it be getting security patches for some period of time? If so you might need a 1.x branch.

I believe we will still support 1.x branch with patches, due to 2.x have a lot of breaking changes.

My two cents:
* The branch `main` will have version `2.x.x` chartversion+appversion, and allow the workflow to auto create releases just like before.
* Create `1.x` branch, will have version `1.x.x` chartverion+appversion. The problem with this is the workflow will only publish anything on `main` to `github-pages`, unless we can support two pages we need to add instructions to use charts with tgz like in ODFE. Add a workflow to create tags for 1.x but not releases.

Thoughts?
@DandyDeveloper @TheAlgo @prudhvigodithi @smlx @mprimeaux ?

Thanks.

smlx:
> I believe we will still support 1.x branch with patches, due to 2.x have a lot of breaking changes.

Ok, thanks for confirming.

> My two cents:
> 
>     * The branch `main` will have version `2.x.x` chartversion+appversion, and allow the workflow to auto create releases just like before.
> 
>     * Create `1.x` branch, will have version `1.x.x` chartverion+appversion.

Agree that this seems like a reasonable solution.

>  The problem with this is the workflow will only publish anything on `main` to `github-pages`, unless we can support two pages we need to add instructions to use charts with tgz like in ODFE. Add a workflow to create tags for 1.x but not releases.

I think you can update the workflow to also publish releases from the `1.x` branch to the same page by adding `1.x` to the list [here](https://github.com/opensearch-project/helm-charts/blob/main/.github/workflows/release.yaml#L7). This might need some testing though.

peterzhuamazon:
> > I believe we will still support 1.x branch with patches, due to 2.x have a lot of breaking changes.
> 
> Ok, thanks for confirming.
> 
> > My two cents:
> > ```
> > * The branch `main` will have version `2.x.x` chartversion+appversion, and allow the workflow to auto create releases just like before.
> > 
> > * Create `1.x` branch, will have version `1.x.x` chartverion+appversion.
> > ```
> 
> Agree that this seems like a reasonable solution.
> 
> > The problem with this is the workflow will only publish anything on `main` to `github-pages`, unless we can support two pages we need to add instructions to use charts with tgz like in ODFE. Add a workflow to create tags for 1.x but not releases.
> 
> I think you can update the workflow to also publish releases from the `1.x` branch to the same page by adding `1.x` to the list [here](https://github.com/opensearch-project/helm-charts/blob/main/.github/workflows/release.yaml#L7). This might need some testing though.

I am not completely sure on that @smlx.
The reason I am saying that is because internally the action will push `a branch` to `gh-pages`. If we push both `main` and `1.x` to `gh-pages` one will override the other.

Also, I dont think github allows you to host two urls for 1 repo.
If we only have one url how can we point to 2 different versions?

And yes we do need some testing on this.
@prudhvigodithi will post another issue soon which includes some findings and proposals based on my initial thoughts, on how to do this the best way soon.

Thanks.

smlx:
> I am not completely sure on that @smlx. The reason I am saying that is because internally the action will push `a branch` to `gh-pages`. If we push both `main` and `1.x` to `gh-pages` one will override the other.

The `chart-releaser-action` _updates_ `index.yaml` on the same `gh-pages` branch, rather than overwriting it.

The `ingress-nginx` repo does this, take a look at their workflow [here](https://github.com/kubernetes/ingress-nginx/blob/main/.github/workflows/helm.yaml), and commits on the `gh-pages` branch which updated the `index.yaml` for a chart release. Here are two example updates: [legacy](https://github.com/kubernetes/ingress-nginx/commit/813c978e6165cecc8ef0380c06ab4cdf106e856b), and [main](https://github.com/kubernetes/ingress-nginx/commit/2872942f74dfe1f15dae71962ef1f9ff1a866877).

Maybe you could ask them for advice on this?

> Also, I dont think github allows you to host two urls for 1 repo. If we only have one url how can we point to 2 different versions?

All the versions are added to the `index.yaml`.

> And yes we do need some testing on this. @prudhvigodithi will post another issue soon which includes some findings and proposals based on my initial thoughts, on how to do this the best way soon.

Sounds good :slightly_smiling_face:

TheAlgo:
>My two cents:
The branch main will have version 2.x.x chartversion+appversion, and allow the workflow to auto create releases just like before.
Create 1.x branch, will have version 1.x.x chartverion+appversion. The problem with this is the workflow will only publish anything on main to github-pages, unless we can support two pages we need to add instructions to use charts with tgz like in ODFE. Add a workflow to create tags for 1.x but not releases.


@peterzhuamazon I partially agree with you. We can have a branch for 1.x, but lets have individual releases for 1.x and 2.x. Also seeing the other comments I see this is feasible and we can maintain 2 branch release same as kubernetes has done for `ingress-nginx`. Going forward I believe this is a good practice to have branches representing major versions (1.x, 2.x, 3.x so on and so forth).

peterzhuamazon:
Thanks everyone, since the issue is up will close this and move the discussion to there.

PR #278, opened 2022-06-24
Add sysctlInit to OpenSearch Chart

### LINKED ISSUE #87: forbidden sysctl: "vm.max_map_count" not whitelisted
Hi,

i've tried to install opensearch via helm, but pods are pending with this message:
forbidden sysctl: "vm.max_map_count" not whitelisted

I've tried to override the security settings, but did not help:
```console
helm upgrade --install opensearch opensearch/opensearch --set sysctl.enabled=true --set podSecurityContext.runAsUser=0 --set securityContext.runAsNonRoot=false >test.yml
```

Im try to migrate from ES, but theirs helm using a bit different method:
they are using a fully privileges initcontainer to set the vm.max_map_count:
```
   initContainers:
154       - name: configure-sysctl
155         securityContext:
156           runAsUser: 0
157           privileged: true
158         image: "docker.elastic.co/elasticsearch/elasticsearch:7.14.0"
159         imagePullPolicy: "IfNotPresent"
160         command: ["sysctl", "-w", "vm.max_map_count=262144"]
161         resources:
```

What would be the best practice? Others does not have this situation?

thanks

amirgo1:
Signed-off-by: amirgo1 <amirgoro1@gmail.com>

### Description
Add option to enable the use of `sysctlInitContainer` to set sysctl vm.max_map_count through privileged `initContainer`.
 
### Issues Resolved
#87 
 
### Check List


peterzhuamazon:
Hi @amirgo1 thanks for contribution,
please resolve the conflicts so we can try to start the workflow and review.
Thanks.

DandyDeveloper:
@peterzhuamazon @amirgo1 Before we approve this as well, we really want to make sure there's a consensus from everyone that this is how we want to do this. 

It's a little worrying that we need to do this via privileged containers that are changing the host OS sysctl settings. 


@smlx @mprimeaux I'll just invite everyone to get a "Yes, lets do this". Maybe thumbs up / Thumbs down?

peterzhuamazon:
Adding @prudhvigodithi to the discussion please see above msg.
Thanks.

charts/opensearch/templates/statefulset.yaml
```diff
@@ -232,6 +232,17 @@ spec:
           - name: "{{ template "opensearch.uname" . }}"
             mountPath: {{ .Values.opensearchHome }}/data
 {{- end }}
+{{- if .Values.sysctlInitContainer }}
```
smlx, line 235: For consistency could we change this to work the same way as the persistence container?
- `sysctlInit.enabled` to enable the functionality
- `sysctlInit.image` and `sysctlInit.imageTag` to control those field.
- default to `busybox:latest` since we don't need any opensearch scripts.

charts/opensearch/templates/statefulset.yaml
```diff
@@ -232,6 +232,17 @@ spec:
...
+      - name: sysctl
+        image: "{{ template "opensearch.dockerRegistry" . }}{{ .Values.image.repository }}:{{ .Values.image.tag | default .Chart.AppVersion }}"
+        imagePullPolicy: "{{ .Values.image.pullPolicy }}"
+        command: ["sysctl", "-w", "vm.max_map_count={{ .Values.sysctlVmMaxMapCount}}"]
```
smlx, line 239: I don't think we should write this every time the init container runs unless it is necessary. Maybe some logic like this to check the value and only write it if required? After all, you don't want to _reduce_ this value if some other service needs it to be even higher.
```
  command:
  - sh
  - -c
  - |
    set -xe
    DESIRED="{{ .Values.sysctl.VmMaxMapCount }}"
    CURRENT=$(sysctl -n vm.max_map_count)
    if [ "$DESIRED" -gt "$CURRENT" ]; then
      sysctl -w vm.max_map_count=$DESIRED
    fi
```

charts/opensearch/README.md
```diff
@@ -96,7 +96,8 @@ helm uninstall my-release
...
 | `sidecarResources`                 | Allows you to set the [resources][] for the sidecar containers in the StatefulSet                                                                                                                                                                         | {}                                              |
-| `sysctlInitContainer`              | Allows you to disable the `sysctlInitContainer` if you are setting sysctl vm.max_map_count` with another method                                                                                                                                        | `enabled: true`                                 |
+| `sysctl`              | Allows you to enable the `sysctl` to set sysctl vm.max_map_count through `securityContext`. This requires privilege.                                                                                                                                         | `enabled: false`                                 |
```
smlx, line 99: I don't think "requires privilege" is enough detail here because the situation is quite complex. Could you add a bit more detail about how this won't work unless you add an argument to the kubelet, and maybe link through to https://kubernetes.io/docs/tasks/administer-cluster/sysctl-cluster/?

charts/opensearch/Chart.yaml
```diff
@@ -15,7 +15,7 @@ type: application
 # This is the chart version. This version number should be incremented each time you make changes
 # to the chart and its templates, including the app version.
 # Versions are expected to follow Semantic Versioning (https://semver.org/)
-version: 2.1.0
+version: 2.1.1
```
smlx, line 18: Because the API (values file) is changing this is at least a minor version bump, rather than a bugfix.

smlx (changes requested):
I'm ok with the concept for this since it's the only practical solution in many k8s environments.

charts/opensearch/README.md
prudhvigodithi: We only need `privileged: true` for `sysctlInitContainer` (that is being pushed via this PR) and not for general `sysctl` and this gets added via  `corev1.PodSecurityContext`, so should we mention this line `This requires privilege` ?
@smlx @amirgo1

peterzhuamazon:
Hi @amirgo1 please resolve the conflicts as well as checking out @smlx review so we can proceed.
Thanks.

charts/opensearch/README.md
amirgo1: @prudhvigodithi I agree with you and think `.Values.sysctl` need to be removed along with `PodSecurityPolicy` as it shouldn't work at all.
vm.max_map_count sysctl is not namespaced, so it cannot be configured through securityContext.

As explained here:
https://github.com/opensearch-project/helm-charts/issues/87#issuecomment-942738258

@smlx - what do you think?

charts/opensearch/Chart.yaml
TheAlgo: Yes this should be a minor version bump. I agree with @smlx

TheAlgo (commented):
I am also okay with this option since I see this as one of the viable options for configuring this property. We have seen some issues with this option when the charts were in the initial phase as well.

charts/opensearch/templates/statefulset.yaml
TheAlgo: Yup this will be un-necessary to do it. To write if only required is a good strategy

charts/opensearch/values.yaml
```diff
@@ -404,12 +404,16 @@ networkPolicy:
...
 fsGroup: ""
 
-## Set optimal sysctl's. This requires privilege. Can be disabled if
+## Set optimal sysctl's through securityContext. This requires privilege. Can be disabled if
 ## the system has already been preconfigured. (Ex: https://www.elastic.co/guide/en/elasticsearch/reference/current/vm-max-map-count.html)
 ## Also see: https://kubernetes.io/docs/tasks/administer-cluster/sysctl-cluster/
 sysctl:
   enabled: false
 
+## Set optimal sysctl's through privileged initContainer.
```
TheAlgo, line 413: Good to add these options in the CI as well so that it is tested everytime : https://github.com/opensearch-project/helm-charts/tree/main/charts/opensearch/ci

charts/opensearch/README.md
smlx: I think that removing `.Values.sysctl` is unrelated to the changes in this PR. It belongs in another PR, and it's a breaking change so would need a major version bump.

peterzhuamazon:
@TheAlgo @smlx @prudhvigodithi please approve this if you are ok with the change, I will resolve the conflicts later.
Thanks.

peterzhuamazon:
@amirgo1 could you check @TheAlgo comments on this?
Thanks.

charts/opensearch/templates/statefulset.yaml
amirgo1: Updated.

charts/opensearch/README.md
amirgo1: I took this out of the documentation (it wasn't there before).

charts/opensearch/values.yaml
amirgo1: I updated all the values files :)

amirgo1:
@peterzhuamazon @TheAlgo my apologies for the delay. I updated the PR

amirgo1:
@TheAlgo, could you please review the changes I made?

charts/opensearch/ci/ci-values.yaml
```diff
@@ -394,12 +394,20 @@ networkPolicy:
...
-## Set optimal sysctl's. This requires privilege. Can be disabled if
+## Set optimal sysctl's through securityContext. This requires privilege. Can be disabled if
 ## the system has already been preconfigured. (Ex: https://www.elastic.co/guide/en/elasticsearch/reference/current/vm-max-map-count.html)
 ## Also see: https://kubernetes.io/docs/tasks/administer-cluster/sysctl-cluster/
 sysctl:
   enabled: false
 
+## Set optimal sysctl's through privileged initContainer.
+sysctlInit:
+  enabled: false
```
TheAlgo, line 405: Can we enable it in one of the CI(s)?

charts/opensearch/CHANGELOG.md
```diff
@@ -13,6 +13,15 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ### Fixed
 ### Security
 ---
+## [2.4.0]
```
TheAlgo, line 16: This needs to become `2.4.2` as there is a PR #279 which is taking up the previous version

charts/opensearch/CHANGELOG.md
```diff
@@ -13,6 +13,15 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
...
+### Added
+- Add option to enable the use of `sysctlInit` to set sysctl vm.max_map_count through privileged `initContainer`. See: [Issue #87](https://github.com/opensearch-project/helm-charts/issues/87)
```
TheAlgo, line 18: Please update the last section of the CHANGELOG which contains the macro for the sections of Unreleased and version specifics

TheAlgo (changes requested):

amirgo1:
@TheAlgo I've made the changes

charts/opensearch/templates/statefulset.yaml
```diff
@@ -236,6 +236,26 @@ spec:
...
+{{- if .Values.sysctlInit.enabled }}
+      - name: sysctl
+        image: "{{ template "opensearch.dockerRegistry" . }}{{ .Values.sysctlInit.image | default "busybox" }}:{{ .Values.sysctlInit.imageTag | default "latest" }}"
+        imagePullPolicy: "{{ .Values.image.pullPolicy }}"
+        command:
+        - sh
+        - -c
+        - |
+          set -xe
+          DESIRED="{{ .Values.sysctl.VmMaxMapCount }}"
```
TheAlgo, line 248: Can we add sample placeholders in `values.yaml` for `vmMaxMapCount`

TheAlgo (changes requested):

TheAlgo:
@smlx @prudhvigodithi Can you guys please review it when free? Thanks

TheAlgo:
@amirgo1 Any update on this PR?

amirgo1:
@TheAlgo Fixed comment and bumped version

TheAlgo (approved):
LGTM!

amirgo1:
Resolved changelog conflict @prudhvigodithi @TheAlgo

TheAlgo:
@prudhvigodithi @DandyDeveloper Please review when free. Thanks

TheAlgo:
@amirgo1 Can you check the lint errors?

peterzhuamazon:
Seems like no update from @amirgo1 I will fix the link for him.

peterzhuamazon:
Thanks @amirgo1 for the contribution and patience, we will merge this PR now.

peterzhuamazon (approved):

peterzhuamazon:
Backport 1.x: https://github.com/opensearch-project/helm-charts/pull/349

Issue #111, opened 2021-10-29
[BUG][opensearch] Parameter discovery.type does not wotk


Sebor:
**Describe the bug**
When I want to setup a single node cluster I set the [options](https://github.com/opensearch-project/helm-charts/blob/main/charts/opensearch/values.yaml#L42) `discovery.type: single-node`. But after that cluster is not started and shows the error:
`setting [cluster.initial_master_nodes] is not allowed when [discovery.type] is set to [single-node]"`

**To Reproduce**
1. Set parameter  `discovery.type` to `single-node` in `values.yaml`
2. Install the chart and look at logs
4. See the above error

**Expected behavior**
Cluster is up and running without errors

**Chart Name**
opensearch

**Host/Environment (please complete the following information):**
 - Helm Version: 3.7.1
 - Kubernetes Version: AWS EKS 1.21.2

Sebor:
Hi @TheAlgo !
Seems it works. But it's not intuitive

TheAlgo:
@Sebor I agree to this. We need to specify these things in our `values.yaml` so that it does not create any confusion. Some settings have been deprecated there it seems and needs to be removed. I will have a look on them.

kunal-parsewar:
Hi , I am also facing the same issue even after making the suggested changes by @TheAlgo .

K8S : 1.22
Virtual Box --> Ubuntu

Also , I am also getting below error 

 suggested_max_allocation_size=256kb, factors={opensearch.unsafe.use_unpooled_allocator=null, g1gc_enabled=true, g1gc_region_size=1mb, heap_size=512mb}]
[2021-12-08T10:30:50,090][INFO ][o.o.d.DiscoveryModule    ] [opensearch-cluster-master-0] using discovery type [zen] and seed hosts providers [settings]
[2021-12-08T10:30:50,441][WARN ][o.o.g.DanglingIndicesState] [opensearch-cluster-master-0] gateway.auto_import_dangling_indices is disabled, dangling indices will not be automatically detected or imported and must be managed manually
[2021-12-08T10:30:50,757][INFO ][o.o.p.h.c.PerformanceAnalyzerConfigAction] [opensearch-cluster-master-0] PerformanceAnalyzer Enabled: false
[2021-12-08T10:30:50,819][INFO ][o.o.n.Node               ] [opensearch-cluster-master-0] initialized
[2021-12-08T10:30:50,820][INFO ][o.o.n.Node               ] [opensearch-cluster-master-0] starting ...
[...]

PR #279, opened 2022-06-29
allow initial_master_nodes to be disabled

hensur:
### Description
[Describe what this change achieves.]
Allows one to not set the `cluster.initial_master_nodes` env var, if a single-node "cluster" is desired.
 
### Issues Resolved
[List any issues this PR will resolve. You should likely open an issue if one does not already exist.]
https://github.com/opensearch-project/helm-charts/issues/111
 
### Check List


prudhvigodithi:
LGTM!
Just curious should we consider to add the replicas to 1 as well in this case  ?
@TheAlgo @DandyDeveloper @peterzhuamazon can you add your thoughts ?

hensur:
@prudhvigodithi Yes, i think that would be useful. Any other amount of replicas wouldn't make sense. I've updated the PR.

prudhvigodithi:
Thanks @hensur.
@TheAlgo @peterzhuamazon @DandyDeveloper @bbarani  if you are ok with this change, should we consider this to backport to 1.x?

prudhvigodithi (approved):

TheAlgo (changes requested):
Can we add this option in the Chart [README](https://github.com/opensearch-project/helm-charts/blob/main/charts/opensearch/README.md)

prudhvigodithi:
Hey @hensur can you please update the README and fix the conflicts ? We can proceed to merge.
Thank you

prudhvigodithi:
Thanks @hensur, @DandyDeveloper can you check?
Thank you

charts/opensearch/values.yaml
```diff
@@ -47,6 +52,7 @@ config:
     network.host: 0.0.0.0
 
     # Setting network.host to a non-loopback address enables the annoying bootstrap checks. "Single-node" mode disables them again.
+    # Requires ".singleNode" to be "true"
     # discovery.type: single-node
```
DandyDeveloper, line 56: Does this have an environment variable that could be used instead?

DandyDeveloper (commented):
I just have a quick Q:

charts/opensearch/values.yaml
hensur: Yes, the `discovery.type` environment variable seems to set this as well. I updated the PR.

charts/opensearch/templates/statefulset.yaml
```diff
@@ -334,6 +338,10 @@ spec:
           value: "{{ .Values.opensearchJavaOpts }}"
         - name: node.roles
           value: "{{ template "opensearch.roles" . }}"
+        {{- if .Values.singleNode }}
+        - name: discovery.type
+          value: "single-node"
+        {{- end }}
```
DandyDeveloper, line 344: Maybe better as: 

```suggestion
          value: {{ if .Values.singleNode }}single-node{{ else }}multi-node{{ end }}
```

Thoughts?

DandyDeveloper (approved):
Fine with the change, just have a suggestion.

charts/opensearch/templates/statefulset.yaml
hensur: Wait, I didn't think to check as I don't have any experience with the opensearch code base, but are you sure there is a `multi-node` discovery type setting?

The sources I've found only list the `zen` type: https://github.com/opensearch-project/OpenSearch/blob/main/server/src/main/java/org/opensearch/discovery/DiscoveryModule.java#L84

hensur: Wouldn't it be best to just omit the env var if we don't set it to `single-node`?
And not make any assumptions about other discovery types in the helm chart?

DandyDeveloper: https://www.elastic.co/guide/en/elasticsearch/reference/current/modules-discovery-settings.html#modules-discovery-settings

Yeah you might be right, maybe this didn't make it over to Opensearch from ES. Weird.

hensur: Weird indeed. Do you think it would be fine to keep the `if` as it is then? Otherwise one would have to update the chart should the default `discovery.type` ever change.

prudhvigodithi: Looks to me like we need `discovery.type` only if we want `single-node`
https://opensearch.org/docs/latest/opensearch/install/docker/#run-the-image

prudhvigodithi:
LGTM!

TheAlgo (approved):
LGTM!

TheAlgo:
@prudhvigodithi @peterzhuamazon Looks like we are good to merge this PR. I have resolved the conflicts as well.

prudhvigodithi:
Thanks @TheAlgo and @hensur  I will merge the PR.

PR #329, opened 2022-09-15
Add healthcheck probes to charts

### LINKED ISSUE #328: [Enhancement][Opensearch/Opensearch-Dashboards] Helm Charts are missing kubernetes health probes
The helm charts don't properly ship with sufficient health checks. While the `opensearch`-chart only lacks a liveness probe, the `opensearch-dashboard`-chart lacks any kind of probes. This causes several problems:
* Deployments of Opensearch Dashboards can't be awaited (since no meaningful readinessProbe is configured)
* Deployments of Opensearch Dashboards are marked as ready, even if they aren't ready to serve traffic yet
* Deployments of both charts arent restarted by Kubernetes when stuck in an error state (only if the application actually crashes)

The missing probes should be added to guarantee availability and reliability of the deployments.

lieberlois:
### Description
This PR introduces the missing healthcheck probes mentioned in #328 . Can be tested using `helm install <release-name> <path-to-chart> --wait` and monitoring the pods during that. The deployment only completes when the respective pods are actually ready to serve traffic.

Note: the values that are configured for the probes are inspired by the official [Opensearch Operator Repository](https://github.com/Opster/opensearch-k8s-operator)
 
### Issues Resolved
#328 
 
### Check List


smlx:
Probes on the largely stateless `opensearch-dashboards` is fine, but putting liveness probes on a slow-starting stateful app like Opensearch needs to be carefully considered, because otherwise you may end up exacerbating problems with the cluster by having Kubernetes constantly restarting nodes. Repeatedly restarting an Opensearch node sounds like a recipe for data loss to me.

Do you have a concrete use-case where having Kubernetes automatically restart the Opensearch nodes would have helped you?

lieberlois:
Hi @smlx , thanks for the quick review.

1. Regarding the livenessProbe for Opensearch: the usecase, this PR is trying to fulfill, is having Opensearch restarted when in an inconsistent state. If Opensearch gets into a state, where it doesn't entirely *crash* but just can't serve requests anymore and doesnt recover (by default, for 20*10 seconds) there isn't much Opensearch itself can do - it has to be restarted. Without that, even a cluster can break down when all of the pods fall into such a state one by one until eventually none of the pods are available anymore. For large clusters, the timeout values can be configured accordingly. However, feel free to disagree, this is just a proposal :)

2. I think this PR is useful for the main branch but also for 1.x versions - should I open a separate PR for that branch? :) not sure whether its possible to contribute to 2.x and backport 1.x at the same time :)

Cheers, Luis

charts/opensearch-dashboards/values.yaml
```diff
@@ -14,6 +14,36 @@ image:
   tag: ""
   pullPolicy: "IfNotPresent"
 
+startupProbe:
```
TheAlgo, line 17: This will not work if OpenSearch is not allowing all traffic right?

lieberlois: The used endpoint is accessible even for anonymous requests.

smlx:
> If Opensearch gets into a state, where it doesn't entirely crash but just can't serve requests anymore and doesnt recover (by default, for 20*10 seconds) there isn't much Opensearch itself can do - it has to be restarted. Without that, even a cluster can break down when all of the pods fall into such a state one by one until eventually none of the pods are available anymore.

Have you experienced this scenario?

sastorsl:
> Have you experienced this scenario?

I have experienced that at least OIDC stopped / failed to work, so having the **possibility** to add a liveness probe is a great addition.

I agree however that it should not be included in `values.yaml` as a standard, and that it should be wrapped in a conditional in `deployment.yaml`

The default values of the readinessProbe should be **very** tolerant in the defaults. I.e. if the backend is gone during patching it is better that OpenSearch Dashboards is available for the duration.

charts/opensearch-dashboards/values.yaml
TheAlgo: Gotcha

TheAlgo:
> 2\. I think this PR is useful for the main branch but also for 1.x versions - should I open a separate PR for that branch? :) not sure whether its possible to contribute to 2.x and backport 1.x at the same time :)

You can raise the backport PR separately once this PR is merged

TheAlgo (changes requested):
@lieberlois Also do you mind fixing the linting issues , I guess there is some trailing space issue here. I will try to test out the chart once post that. Thanks

lieberlois:
@TheAlgo Thank you, I'll try to get to it this evening :) thanks for the review!

lieberlois:
@sastorsl I changed the livenessProbe to be an optional parameter in the values.yaml :) feel free to suggest any default values, however I think the chosen values aren't too bad, since they are being used in the official [Operator Project](https://github.com/Opster/opensearch-k8s-operator) :)

Thanks for your review! :)

sastorsl:
> @sastorsl I changed the livenessProbe to be an optional parameter in the values.yaml :) feel free to suggest any default values, however I think the chosen values aren't too bad, since they are being used in the official [Operator Project](https://github.com/Opster/opensearch-k8s-operator) :)
> 
> Thanks for your review! :)

Re opensearch it looks good.

I think the `livenessProbe` should be optional for OpenSearch Dashboards aswell.

lieberlois:
Any updates on this @TheAlgo :) ? I think all the requested changes are implemented in this PR ^^

TheAlgo:
> Any updates on this @TheAlgo :) ? I think all the requested changes are implemented in this PR ^^

Just 1 last small thing, can you add this feature in the opensearch-dashboards [README](https://github.com/opensearch-project/helm-charts/blob/main/charts/opensearch-dashboards/README.md) as well?

lieberlois:
@TheAlgo I added the features to the README files. 

Regarding the CI: the issue is, that opensearch-dashboards cant properly start, when no opensearch instance is found. The log clearly shows the following error message:

```json
{"type":"log","@timestamp":"2022-09-27T07:25:51Z","tags":["error","opensearch","data"],"pid":1,"message":"[ConnectionError]: getaddrinfo ENOTFOUND opensearch-cluster-master"}
```

How do you want to go about this? Making all probes for opensearch dashboards would fix the CI, but then, a proper startup isnt really depicted in the CI

Thanks in advance 👍 
Luis

sastorsl:
> Regarding the CI: the issue is, that opensearch-dashboards cant properly start, when no opensearch instance is found. The log clearly shows the following error message:

There is another issue discussing exactly this, I don't recall if it was in `helm-charts` or in the OpenSearch Dashboards branch.

For now - until any change occurs - I don't see how you can get CI as long as OpenSearch Dashboards don't start without a functioning OpenSearch backend.

IMHO it **should** start and give the user a proper feedback, respond to health checks, etc, but that's another issue entirely...

smlx:
If dashboards liveness probes reports that the pod is unhealthy because the opensearch it is configured to talk to is not available, that's a classic cascading failure case. There is no need to restart dashboards pods just because opensearch is not available. The liveness probe health check should use an endpoint that does not rely on external dependencies.

lieberlois:
Very good point @smlx , the current configuration was taken from the [official operator repository](https://github.com/Opster/opensearch-k8s-operator/blob/8b1d3f88718d17c6a7b137bca437788e44aac5a1/opensearch-operator/pkg/builders/dashboards.go#L73). Any ideas on a more suitable liveness probe? :)

sastorsl:
Letting OpenSearch Dashboards run without the backend dependency is something which must be solved in that project. I created an issue to that effect which I linked to the issue - https://github.com/opensearch-project/OpenSearch-Dashboards/issues/2422

@lieberlois I think the default lifecycle probe should be either _only_ a TCP check, or disabled and left to the individual user to add / configure.

TheAlgo:
There was some discussion around it 1 year back and from then it did not grab much attention

PR Link : https://github.com/opensearch-project/helm-charts/pull/117

lieberlois:
I quite like @sastorsl 's idea to just use a tcp socket. Just tested it, works properly with a standalone opensearch dashboards deployment and can be overridden for deployments that aim for opensearch dashboards unavailability when opensearch itself is down :) I think we cant go any further than this until the opensearch dashboards application itself can handle a missing opensearch cluster.

Thanks for all of your reviews and advice! @TheAlgo are you fine with this solution?

TheAlgo:
> I quite like @sastorsl 's idea to just use a tcp socket. Just tested it, works properly with a standalone opensearch dashboards deployment and can be overridden for deployments that aim for opensearch dashboards unavailability when opensearch itself is down :) I think we cant go any further than this until the opensearch dashboards application itself can handle a missing opensearch cluster.
> 
> Thanks for all of your reviews and advice! @TheAlgo are you fine with this solution?

Yes I am fine. We did the same even in OpenSearch

PR : https://github.com/opensearch-project/helm-charts/pull/172

lieberlois:
Great! If you don't have any further requests, LGMT 🚀

TheAlgo (approved):

TheAlgo:
@lieberlois I have approved the PR but you need to update the CHANGELOG to add the unreleased section. 

Ref : https://github.com/opensearch-project/helm-charts/blob/main/CONTRIBUTING.md

peterzhuamazon:
Adding @prudhvigodithi to take a look and I am fine with this PR.
Like @TheAlgo mentioned we need to update the changelog.

Thanks.

lieberlois:
@TheAlgo @peterzhuamazon Thanks :) i updated the changelog 👍

peterzhuamazon (approved):

peterzhuamazon:
> @TheAlgo @peterzhuamazon Thanks :) i updated the changelog 👍

Thanks @lieberlois are you able to backport it to 1.x branch?
Thanks.

lieberlois:
@peterzhuamazon I will, i was told to do that in a separate PR ☺️👍

peterzhuamazon:
> @peterzhuamazon I will, i was told to do that in a separate PR ☺️👍

Yes @lieberlois please backport once this one is merged.
Thanks!

lieberlois:
LGTM then everybody! 🚀

prudhvigodithi (approved):

peterzhuamazon:
@lieberlois It is merged now thanks for all the contributions!
Let me know when the backport is ready.

Thanks.

lieberlois:
Thanks everybody!

@peterzhuamazon You can expect it on tuesday, I'm on a small vacation until then ☺️

Issue #339, opened 2022-10-20
[Enhancement][opensearch] allow change the volumeClaimTemplate name


retoo:
**Is your feature request related to a problem? Please describe.**
The current volumeClaimTemplate name is rather long, currently the name `opensearch.uname`being used. The problem is, that this leads to really long PVC names, i.e. if your `clusterName` is `opensearch-test`then your `volumeClaimTemplate`will be `opensearch-test-master` which leads to the pvc name `opensearch-test-master-opensearch-test-master-0`

**Describe the solution you'd like**
The name itself probably cannot be changed without breaking old installations. Therefore we should make the name itself configurable.

smlx:
Could you explain a bit why the current naming scheme is causing problems for you?

retoo:
It just creates super long PVC names. Lets say I name my cluster `opensearch-myapp-integration`, this leads to both the claim-template and the statefulset to be called: `opensearch-myapp-integration-master` (the groupname `master`gets appended).

When a statefulset scales up, it looks for a claim named: `$CLAIM_TEMPLATE_NAME-$STATEFULSET_NAME-{0..N}`

Meaning, it will create a PVC named: `opensearch-myapp-integration-master-opensearch-myapp-integration-master-0` which is both ridiculous long (74 chars) ;) 

The claimTemplate name just need to be unique within one statefulset, other vendors name the template `data`, which for example would lead to `data-$CLUSTER_NAME-master-0`.

Unfortunately it is not possible to change the pvc name now, it probably would break old setup, hence the override for people like me with long names ;).

**tldr** Ugly long names, possible break k8s identifier limit.

TheAlgo:
If this is something which is really impacting users of the chart, then I don't have any strong opinions for this change. We can keep the default value and give a facility for configuring it.

cpockrandt:
I agree, the PVC names are very long, which causes them to be truncated in our monitoring/logging dashboards. A way to overwrite the names would be very helpful in my opinion.

PR #340, opened 2022-10-20
Provide a way to override the volumeClaimTemplate name

retoo:
### Description
The current volumeClaimTemplate name is rather long and leads to long PVC names. This PR introduces a new value `persistence.volumeClaimTemplateNameOverride`which allows the user to change the name. By default the value is not defined and the current name remains in use. 

 
### Issues Resolved
- #339
 
### Check List

- [ ] Helm chart version bumped


peterzhuamazon:
Hi @retoo thanks for the contribution.
Is this something we can add for dashboards as well?

@TheAlgo @prudhvigodithi 

Thanks.

retoo:
@peterzhuamazon yes, that should be possible. I try to look into it next weekend.

peterzhuamazon:
Hi @DandyDeveloper @TheAlgo @prudhvigodithi please take a look and review.

Thanks.

DandyDeveloper (changes requested):
Would it not be more ideal to instead implement a method for overwriting the entire volumeClaimTemplate? This seems like overload just for a name.

```
volumeClaimTemplates:
{{- range .Values.persistence.volumeClaimTemplates }}
- {{ toYaml (tpl .name $root) }} 
{{- end }}
```

Then in values:

```
persistence:
  volumeClaimTemplates:
    - name: {{ include "xyz.123" . }} 
       ...
 ```

Having an explicit override just for a name feels like bloat. 

With this you also give people the chance to add new volumes if they want for plugins / certs / whatever they want, which they can consume with initContainers. 

It's above and beyond the scope of your specific need, but I think we might as well at least implement something others can use rather than this niche use case described.

prudhvigodithi:
Hey @retoo can you please check @DandyDeveloper comment and see if that is a valid scenario ?
Thank you

prudhvigodithi:
Hey @retoo closing this, please re-open if required.
Thank you

PR #503, opened 2023-12-07
Updates HELM configuration to adapt to changes in demo admin user setup and bumps appVersion to 2.12.0

DarshitChanpura:
### Description
This change adapts to the change in security plugin's demo installation script which now requires admin password to be setup when installing demo configuration. Without this password, the cluster will not spin up.
 
### Issues Resolved
 - Related to https://github.com/opensearch-project/security/issues/3711
 
### Check List

- [ ] Helm chart version bumped
- [ ] Helm chart `CHANGELOG.md` updated to reflect change


DarshitChanpura:
@prudhvigodithi to add changes to support for `initialAdminPassword.txt`

charts/opensearch/values.yaml
```diff
@@ -104,6 +104,10 @@ config:
 extraEnvs: []
 #  - name: MY_ENVIRONMENT_VAR
 #    value: the_value_goes_here
+# 2.12 onwards a custom strong password needs to be provided in order to setup demo admin user.
+# Cluster will not spin-up without this unless demo config install is disabled.
+#  - name: OPENSEARCH_INITIAL_ADMIN_PASSWORD
+#    value: <strong-password>
```
ruanyl, line 110: I normally configure demo users(including `admin`) via `internal_users.yml` which password is hashed. Do I still need this env variable? Should it skip demo user initialization if `internal_users.yml` already contains admin password?

charts/opensearch/values.yaml
```diff
@@ -104,6 +104,10 @@ config:
 extraEnvs: []
 #  - name: MY_ENVIRONMENT_VAR
 #    value: the_value_goes_here
+# 2.12 onwards a custom strong password needs to be provided in order to setup demo admin user.
+# Cluster will not spin-up without this unless demo config install is disabled.
```
ruanyl, line 108: Shall we highlight this in the README of this repo? in [Notes About Default Installation](https://github.com/opensearch-project/helm-charts?tab=readme-ov-file#notes-about-default-installation)

charts/opensearch/values.yaml
DarshitChanpura: > Do I still need this env variable?
Yes, for demo `admin` user setup this env variable will still be required.

> Should it skip demo user initialization if internal_users.yml already contained admin user?

If the security was already configured (i.e. opensearch.yml ) then the demo script will not execute. This variable is needed for fresh installs only.

charts/opensearch/values.yaml
DarshitChanpura: Good call-out. I will push an update

charts/opensearch/values.yaml
ruanyl: > If the security was already configured (i.e. opensearch.yml ) then the demo script will not execute. This variable is needed for fresh installs only.

If I just want the script to setup demo certificates for me as I configured the demo users myself, can I skip this env variable?

DarshitChanpura: no. If you run demo script, this variable is needed. It is not required only when running a setup without demo configuration.

DarshitChanpura:
@prudhvigodithi @ruanyl Can I get more reviews on this?

charts/opensearch/Chart.yaml
```diff
@@ -15,7 +15,7 @@ type: application
 # This is the chart version. This version number should be incremented each time you make changes
 # to the chart and its templates, including the app version.
 # Versions are expected to follow Semantic Versioning (https://semver.org/)
-version: 2.17.2
+version: 2.17.3
```
peterzhuamazon, line 18: This is at least a minor change, not patch.

charts/opensearch/values.yaml
```diff
@@ -104,6 +104,10 @@ config:
...
+#  - name: OPENSEARCH_INITIAL_ADMIN_PASSWORD
```
peterzhuamazon, line 109: Can we at least output some error msg for people to use to follow?
Instead of failing out of nowhere and need to check readme?

charts/opensearch/values.yaml
```diff
@@ -104,6 +104,10 @@ config:
 extraEnvs: []
 #  - name: MY_ENVIRONMENT_VAR
 #    value: the_value_goes_here
+# 2.12 onwards a custom strong password needs to be provided in order to setup demo admin user.
```
peterzhuamazon, line 107: This should be clearly state as chartVersion 2.18.0 and the appVersion 2.12.0 onward, or it will confuse people on this.

README.md
```diff
@@ -71,6 +71,15 @@ extraEnvs:
     value: "true"
 ```
 
+OpenSearch version 2.12 and above require a custom strong password to be provided in order to setup demo admin user. Without this password the cluster would not spin up, unless demo config install is disabled.
```
peterzhuamazon, line 74: This should be highlighted as ## Breaking Changes or something similar, or at the top of the readme or users will miss it easily.

peterzhuamazon (changes requested):

charts/opensearch/Chart.yaml
peterzhuamazon: Use 2.18.0.

peterzhuamazon: #504 will use 2.18.1

charts/opensearch/values.yaml
DarshitChanpura: I will need your help adding that.

peterzhuamazon: After discussing with @DarshitChanpura and @derek-ho seems like we cannot do too much on the log, as kubectl will show pods status as ERROR, and user can just check kubectl logs for the exact comment here:
https://github.com/opensearch-project/security/blob/c06365ca94c9a1d15e85b578a1ae48168bf0bca7/src/main/java/org/opensearch/security/tools/democonfig/SecuritySettingsConfigurer.java#L155-L157

However, I wonder if we can change the commend of helm install, to reflect this env var requirement directly in the commend, so people are aware that we need to set such env var in order to deploy the pods successfully.
https://github.com/opensearch-project/security/blob/c06365ca94c9a1d15e85b578a1ae48168bf0bca7/src/main/java/org/opensearch/security/tools/democonfig/SecuritySettingsConfigurer.java#L155-L157

DarshitChanpura: That is not an option here, since we'll have to pass a default plain-text password in values.yml anyway if we want to allow users to set via --set. We are trying to move away from default passwords and hence this is not be the best path forward.

README.md
```diff
@@ -1,9 +1,12 @@
 <img src="https://opensearch.org/assets/brand/SVG/Logo/opensearch_logo_default.svg" height="64px"/>
 
-- [OpenSearch Project Helm-Charts](#helm-charts)
+- [Helm-Charts](#helm-charts)
```
DarshitChanpura, line 3: this change was auto done by VS code, and we should keep it

peterzhuamazon (commented):
Please also add a CHANGELOG entry for opensearch on 2.18.0.

README.md
```diff
@@ -58,6 +62,16 @@ helm install my-deployment opensearch/<chart name>
 
 Please see the `README.md` in the [OpenSearch](charts/opensearch) and [OpenSearch Dashboards](charts/opensearch-dashboards) directories for installation instructions.
 
+#### Breaking Change
```
peterzhuamazon, line 65: Why use `####` instead of `###`?

charts/opensearch/CHANGELOG.md
```diff
@@ -12,7 +12,16 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ### Removed
 ### Fixed
 ### Security
-
+---
+## [2.18.0]
+### Added
+### Breaking
+ - Requires an initial admin password to be setup. Refer this github issue: https://github.com/opensearch-project/security/issues/3622
```
peterzhuamazon, line 19: Please add `since OpenSearch AppVersion 2.12.0` after `to be setup`.

peterzhuamazon (changes requested):

README.md
DarshitChanpura: updated.

peterzhuamazon (approved):

DarshitChanpura:
@prudhvigodithi Could you please review this?

DarshitChanpura:
@prudhvigodithi Could you please add 1 more review?

prudhvigodithi:
Thanks @DarshitChanpura, this PR should be merged after `2.12.0` release because it clearly said `2.18.0`  requires `OPENSEARCH_INITIAL_ADMIN_PASSWORD` which is not the case if we merge this PR today. So I would suggest to hold on this PR and merge it with 2.12.0 release. Also since this `OPENSEARCH_INITIAL_ADMIN_PASSWORD` is a mandatory option for new users, the `extraEnvs` are optional so this might get lost. 

I would suggest to use a new parameter as `initialAdminPassword` in [values.yaml](https://github.com/opensearch-project/helm-charts/blob/main/charts/opensearch/values.yaml) (example https://github.com/opensearch-project/helm-charts/blob/main/charts/opensearch/values.yaml#L3) and backend in statefulset pass this as ENV.

With your change can you please test adding a `keystore: []` and enabling `masterTerminationFix`? because if uses the password with these options.

Thanks

DarshitChanpura:
> I would suggest to use a new parameter as initialAdminPassword in [values.yaml](https://github.com/opensearch-project/helm-charts/blob/main/charts/opensearch/values.yaml) (example https://github.com/opensearch-project/helm-charts/blob/main/charts/opensearch/values.yaml#L3) and backend in statefulset pass this as ENV.
With your change can you please test adding a keystore: [] and enabling masterTerminationFix? because if uses the password with these options.

@prudhvigodithi I'm unfamiliar with HELM and hence slightly confused at what the ask is here. We are not supplying a default password. User will always have to pass the password. Our intention is to fail cluster startup if no password was supplied. By adding a new parameter "initialAdminPassword" we are essentially supplying a default value. We are moving away from default hardcoded password for admin.

prudhvigodithi:
Thanks @DarshitChanpura for your contribution, there are some more changes required, I'm closing this PR in favor of https://github.com/opensearch-project/helm-charts/pull/518.

PR #504, opened 2023-12-07
Comment out hard coded opensearch.yml

derek-ho:
### Description
This PR stops feeding in a hard coded opensearch.yml into the pods. We are expecting to generate this output from running the install demo configuration script from the security plugin side.
 
### Issues Resolved
[List any issues this PR will resolve. You should likely open an issue if one does not already exist.]
 
### Check List


charts/opensearch/values.yaml
```diff
@@ -31,71 +31,71 @@ global:
...
-    # Bind to all interfaces because we don't know what IP address Docker will assign to us.
-    network.host: 0.0.0.0
-
-    # Setting network.host to a non-loopback address enables the annoying bootstrap checks. "Single-node" mode disables them again.
-    # Implicitly done if ".singleNode" is set to "true".
-    # discovery.type: single-node
-
-    # Start OpenSearch Security Demo Configuration
-    # WARNING: revise all the lines below before you go into production
-    plugins:
```
derek-ho, line 60: This file should be generated via the demo configuration script, if it is hard coded we will need to maintain this in two places.

TheAlgo (commented):
But this sample yaml is very helpful for anyone starting opensearch right?

smlx:
I don't understand this change. This is a helm chart, they are configured via values files.. how are you generating your configuration otherwise?

DarshitChanpura:
@smlx @TheAlgo When an OpenSearch image is a downloaded and cluster is spun up, the Security plugin runs install demo configuration by default as part of install. During this phase, opensearch.yml is loaded with security related values. Hence we do not need to hardcode opensearch.yml. Feel free to experiment and lmk if otherwise.

charts/opensearch/CHANGELOG.md
```diff
@@ -13,6 +13,15 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ### Fixed
 ### Security
 
+---
+## [2.17.3]
```
peterzhuamazon, line 17: Use 2.18.1 as I put #503 as #2.18.0.

charts/opensearch/Chart.yaml
```diff
@@ -15,7 +15,7 @@ type: application
 # This is the chart version. This version number should be incremented each time you make changes
 # to the chart and its templates, including the app version.
 # Versions are expected to follow Semantic Versioning (https://semver.org/)
-version: 2.17.2
+version: 2.17.3
```
peterzhuamazon, line 18: 2.18.1

charts/opensearch/CHANGELOG.md
```diff
@@ -13,6 +13,15 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
...
+### Added
+### Changed
+### Deprecated
+### Removed
+- Removed hard coded opensearch.yml
```
peterzhuamazon, line 22: This is a breaking change, please be clear in the changelog.

peterzhuamazon (changes requested):

peterzhuamazon (approved):

DarshitChanpura:
@prudhvigodithi Could you please review this?

DarshitChanpura:
@prudhvigodithi Could you please add 1 more review?

prudhvigodithi:
Hey Just tested this and works fine with removing the `.Values.config` , basically when user adds the `opensearch.yml` through `.Values.config`, it gets added to `/usr/share/opensearch/config/opensearch.yml` overriding the default. Also one advantage is next if there is any change in  `opensearch.yml` and added through `.Values.config`, the pods get auto restarted and loads the new values from `opensearch.yml`.  So @DarshitChanpura and @derek-ho any specific reason we are removing, we can leave it as well right ?

DarshitChanpura:
> So @DarshitChanpura and @derek-ho any specific reason we are removing, we can leave it as well right ?

From what I understand, the opensearch.yml always gets created when demo configuration script is run (which is default). If in case the user decides to pass in custom configuration, they can always pass it in via .values.yml. In either of these two cases, the hardcoded opensearch.yml serves no purpose. Which is why we are removing this? @derek-ho Please add if I missed anything here.

smlx:
I don't know if it is on the roadmap of the Opensearch image to avoid generating the demo config at runtime but IMO it should be. Generating config at runtime has the following consequences for docker and k8s:

* it slows down container startup
* root filesystem can't be read only
* can't run the container as an arbitrary user (has to have permissions to write the config file)

IMO it makes more sense to aim for immutable images with defined injection points for configuration rather than generating config at runtime. Therefore I think this PR is a move in the wrong direction.

prudhvigodithi:
Instead of commenting out `opensearch.yml`, I have raised a PR https://github.com/opensearch-project/helm-charts/pull/516, that address configMap Read-only file system error when passed via `.Values.config`, with this the user can either use the default `opensearch.yml` or pass the `opensearch.yml` via `.Values.config`. The file will be created as `-rw-r--r-- 1 opensearch opensearch 3646 Feb  7 03:22 opensearch.yml`, having RW access within the cluster,  please check.

@TheAlgo @DarshitChanpura @peterzhuamazon @bbarani @derek-ho @smlx

derek-ho:
I can close this out, since I tested out the alternative also works and it seems the direction of the community is not towards removing this.

DarshitChanpura:
This can be closed out since the changes in #516 addresses the read-only issue when writing demo configuration.

PR #619, opened 2024-11-09
feat: Add ServiceMonitor for data-prepper

darnaut:
### Description
Adds `ServiceMonitor` resource for monitoring with Prometheus.
 
### Issues Resolved
N/A
 
### Check List

### DIFF
--- a/charts/data-prepper/CHANGELOG.md
+++ b/charts/data-prepper/CHANGELOG.md
@@ -4,7 +4,29 @@ All notable changes to this project will be documented in this file.
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.1.0/),
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
+---
 ## [Unreleased]
 ### Added
+### Changed
+### Deprecated
+### Removed
+### Fixed
+### Security
+---
+## [0.1.1]
+### Added
+- ServiceMonitor resource for Prometheus monitoring
+### Changed
+### Deprecated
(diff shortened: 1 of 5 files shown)


peterzhuamazon:
Hi @TheAlgo @prudhvigodithi could you take a look on this?

Thanks.

sgorshkov85:
Hello. 
 Currently(with what to have: https://github.com/opensearch-project/helm-charts/blob/data-prepper-0.1.0/charts/data-prepper/templates/deployment.yaml#L49-L57), it makes sense to use PodMonitor instead. 
 OR - adjust https://github.com/opensearch-project/helm-charts/blob/data-prepper-0.1.0/charts/data-prepper/templates/service.yaml#L9-L15 to add `server`(port 4900) to the list, so the approach with service monitor will also work.

darnaut:
@sgorshkov85 For context, the choices made here were intentional:

1. `ServiceMonitor` is typically the recommend approach by Prometheus and also works better here given that there's a service available and it will only match the endpoints available in the service (e.g. it won't scrape pods removed from the service)
2.  The port is not enabled by default because the `ServiceMonitor` itself is not enabled by default. It's left up to the Helm chart user to enable the required port.

darnaut:
@sgorshkov85 if there's some issue for you, please submit a PR fixing the duplicate port issue. this would be a problem regardless of this change (e.g. can't expose the server port in the service).

prudhvigodithi:
Adding @eyenx to please take a look.
@getsaurabh02

PR #634, opened 2024-12-17
fix(chart/data-prepper) Remove demo pipeline from values, add explicit demo value

### LINKED ISSUE #624: [BUG][data-prepper] Chart values should not have a default pipeline defined
**Describe the bug**
The default values file has a demo pipeine defined. Even if you in your own values file define a `my-pipeline`, the demo `simple-sample-pipeline` will be deployed alongside it.

**To Reproduce**
1. Create a simple `myvalues.yaml` file with a `pipelineConfig.config.my-pipeline` defined
2. Install chart with `-f myvalues.yaml`
3. Look at the `foo-data-prepper-pipeline` secret - it contains both pipelines.

**Expected behavior**
You'd expect that the pipeline you configure is the only one installed.

You can achieve a blank slate today by using the `existsingSecret` property, but for simple pipelines the convenience of defining it in values file should achieve the same.

**Chart Name**
data-prepper

**Screenshots**
Skjermbilde 2024-12-04 kl  11 21 49

**Host/Environment (please complete the following information):**
 - Helm Version: 3.16.3
 - Kubernetes Version: 1.30.2

**Additional context**
Proposed fix is to ship chart without a configured pipeline. Alternatively use a `noop` pipeline that does nothing, to make the prepper not crash. Instead, ship an `examples/` folder and move the `simple-sample-pipeline` example to a separate values file in that folder so users can do `helm install myprepper opensearch/data-prepper -f example-values.yaml` to test.

janhoy:
### Description
This PR comments demo pipline from values, and also removes the alread-commented complex pipeline example, refering to documentation instead.

Now, when installing the chart with defaults only, the install will fail with this message:

```
helm.go:86: 2024-12-17 15:21:55.866789 +0100 CET m=+0.033838001 [debug] execution error at (data-prepper/templates/deployment.yaml:98:27): A valid pipeline configuration is required! Please set one of:
pipelineConfig.existingSecret - for your custom pipeline secret
pipelineConfig.enabled - for inline configuration in values.yaml or
pipelineConfig.demoPipeline=true - for a sample demo pipeline using random source and stdout sink.
```

If you instead install the chart with `--set pipelineConfig.demoPipeline=true`, the behavior will be identical to current chart behavior, and we also print an informative message in stdout about what pipeline configuration is in use:

```
...
The demo pipeline configuration is enabled, using the `random` source and `stdout` sink.
```

### Issues Resolved
Fixes #624
 
### Check List


prudhvigodithi:
Adding @SergK to please take a look as well.
@getsaurabh02

prudhvigodithi:
Thanks @janhoy I like the approach to use `--set pipelineConfig.demoPipeline=true`, can you please fix the failing the CI tests?

DandyDeveloper:
@janhoy I've pushed out the updates to hopefully get the linter passing. 

FYI, by default this would be a breaking change for anybody upgrading. Specifically because the default stance of the `values.yaml` assumes neither the demopipeline or user inputted pipeline has been enabled. 

My recommendation is we keep existing behaviours and have the `demoPipeline` enabled by default.

janhoy:
> My recommendation is we keep existing behaviours and have the `demoPipeline` enabled by default.

Yes, it was breaking change. I modified the behavior so that the demo pipeline is installed automatically only when no pipeline-secret or inline pipeline is enabled. The `demoPipeline` value defaults to "" which causes this auto mode, but it can still be overridden with `true` to explicitly install demo pipline (in which case any other configured pipeline will NOT be installed, or `false` in which case helm install will fail if no other pipeline is configured.

I do not see a benefit in preserving the possibility of ADDING a demo pipeline to an existing one, it would also be quite complicated. The "simple-sample-pipeline" is commented in values.yaml, so users can achieve this by modifying their own pipeline.

DandyDeveloper (approved):
LGTM. Agree with the above.

janhoy:
Thanks for approving, can we get it merged? 😉

DandyDeveloper:
I'm gonna merge since I can't see a reason why we don't want this.

PR #654, opened 2025-02-21
update statefulset to inlcude verbose PVC reference

ritzk:
### Description
The issue is that the verbose reference is added once ArgoCD applies the chart to cluster. This results in ArgoCD thinking there is a diff, when there is none.  This can be solved by adding a patch which adds the verbose reference.
 
### Issues Resolved
n/a
 
### Check List

- [X ] Helm chart version bumped
- [ x] Helm chart `CHANGELOG.md` updated to reflect change


ritzk:
> @ritzk Please do DCO Checks and check the linting output.

```
$ helm lint  .
==> Linting .
[INFO] Chart.yaml: icon is recommended

1 chart(s) linted, 0 chart(s) failed
```

DandyDeveloper:
@ritzk One thing you will wanna test, can you install the original, then upgrade with your changes. Can you confirm if this "breaks" because the STS isn't immutable? This will cause some real problems because people would effectively need to remove the STS, then apply these updates which could bring multiple nodes down on an upgrade if forced.

DandyDeveloper:
@ritzk Sorry for the delay, I mean we need to test an install from the current chart version, then upgrade to yours. Because the change in spec MAY complain that the `StatefulSet` resource is immutable and cannot be changed.

This will mean people upgrading will all break and have to forcefully remove and add the statefulset. This could be bad! 

Can you check this?

ritzk:
> @ritzk Sorry for the delay, I mean we need to test an install from the current chart version, then upgrade to yours. Because the change in spec MAY complain that the `StatefulSet` resource is immutable and cannot be changed.
> 
> This will mean people upgrading will all break and have to forcefully remove and add the statefulset. This could be bad!
> 
> Can you check this?

Works fine.

install
```
opensearch ±|main|→ git remote  -v
origin	https://github.com/opensearch-project/helm-charts.git (fetch)
origin	https://github.com/opensearch-project/helm-charts.git (push)

opensearch ±|main|→ helm upgrade --install opensearch -n opensearch --create-namespace .
Release "opensearch" does not exist. Installing it now.
NAME: opensearch
LAST DEPLOYED: Sat Mar  8 14:16:41 2025
NAMESPACE: opensearch
STATUS: deployed
REVISION: 1
TEST SUITE: None
NOTES:
Watch all cluster members come up.
```

upgrade
```
opensearch ±|argo-sync ✗|→ git remote  -v
origin	git@github.com:ritzk/opensearch-helm-charts.git (fetch)
origin	git@github.com:ritzk/opensearch-helm-charts.git (push)

opensearch ±|argo-sync ✗|→ helm upgrade --install opensearch -n opensearch --create-namespace . 
Release "opensearch" has been upgraded. Happy Helming!
NAME: opensearch
LAST DEPLOYED: Sat Mar  8 14:24:46 2025
NAMESPACE: opensearch
STATUS: deployed
REVISION: 2
TEST SUITE: None
NOTES:
Watch all cluster members come up.
  $ kubectl get pods --namespace=opensearch -l app.kubernetes.io/component=opensearch-cluster-master -w

```

PR #720, opened 2025-12-19
Allow override of core Opensearch container env variables

briend:
### Description
Allow complete override of `env` for the opensearch container. Some users would like to customize env vars such as `node.name`.  The `extraEnvs` key is not sufficient as it can introduce [duplicate keys and patch errors with Helm v4](https://github.com/helm/helm/issues/31529 ), and is generally not a best practice.
 
### Issues Resolved
- https://github.com/opensearch-project/helm-charts/issues/721
- https://github.com/helm/helm/issues/31529 

### Check List

### DIFF
--- a/charts/opensearch/CHANGELOG.md
+++ b/charts/opensearch/CHANGELOG.md
@@ -14,6 +14,15 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ### Fixed
 ### Security
 ---
+## [3.4.1]
+### Added
+- envTpl values key to allow overriding core opensearch container env
+### Changed
+### Deprecated
+### Removed
+### Fixed
+### Security
+---
 ## [3.4.0]
 ### Added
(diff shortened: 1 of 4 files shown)


briend:
I don't know if this is the best solution, it might be a bit heavy-handed to have string templates in the Values.yaml file.  Our use case is that we want this in our `opensearch` container `env`; so we need to have `spec.nodeName` defined as a variable first.  This worked fine in `extraEnvs` until the recent helm upgrade, because there are duplicated keys.

```
env:
- name: NODE_NAME
  valueFrom:
    fieldRef:
      apiVersion: v1
      fieldPath: spec.nodeName
- name: node.name
  value: '${HOSTNAME}-${NODE_NAME}'  # we want data nodes to include the actual k8s nodeName
```

briend:
We've decided we don't really need to customize those core variables like `node.name`, so I'll close this and the PR.  The issue remains but I doubt it will affect anyone

PR #742, opened 2026-05-22
Pin GitHub Actions to commit SHAs

Divyaasm:
### Description

Pin all GitHub Action tag references to their corresponding commit SHAs.

Tags are mutable references that can be force-pushed to point to different commits, making them vulnerable to supply chain attacks. Commit SHAs are immutable and guarantee that the exact reviewed code is executed in CI workflows. This change pins all third-party actions to their current commit SHAs to prevent potential tampering.

### DIFF
--- a/.github/workflows/github-merit-badger.yml
+++ b/.github/workflows/github-merit-badger.yml
@@ -11,7 +11,7 @@ jobs:
     permissions:
       pull-requests: write
     steps:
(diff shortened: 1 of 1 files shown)


gaiksaya:
@Divyaasm Looks like multiple other workflows in this repo needs to pin the SHA as well

peterzhuamazon:
https://github.com/opensearch-project/helm-charts/pull/743

PR #753, opened 2026-07-30
Handle release namespace

andrejvanderzee:
### Description
Fix charts for kustomization v5.8.0
 
### Issues Resolved
namespace is not set on Kubernetes resources since v5.8.0.

###
For reference:
https://github.com/kubernetes-sigs/kustomize/issues/6058


anthops:
Keen for this 🤞

